PR #2, opened 2026-05-28
Multi-host tabs with dynamic add/remove

srid:
**drishti can now monitor several hosts at once from one parent server**, with a tab strip across the top, a live connection-state dot on each tab, and `+ add host` / `×` controls for runtime add/remove. The host set persists to `$XDG_STATE_HOME/drishti/hosts.json` so UI changes survive a restart — `nix run github:srid/drishti` with no args restores the last session.

Design discussion (including the rejected schema-keyed alternative) lives in `talk-multi-host.html` at the repo root; the diff reflects the post-review proposal.

### How it fits together

```
Browser (one WS per host)
   │
   │  /rpc/ws?host=<id>            /rpc/ws?host=__admin__
   ▼                                       │
Parent (HostRegistry)         Admin router (hosts collection,
   │  ssh stdio × N            add/remove procedures)
   ▼
drishti-agent × N
```

_Host identity lives only at the transport layer — the per-host `surface` schema (system / processes / cpuCores / connection) is unchanged. A separate admin surface carries the host set as `Collection<string, HostEntry>` plus `hosts.add` / `hosts.remove` procedures; the tab strip subscribes to that collection and the `+` / `×` buttons call the procedures._

### What you get

- **CLI** — `drishti host1 host2 host3` (positional, parsed by `cleye` to match kolu's CLI style). Default `["localhost"]` when none.
- **Tab strip** — chip per host with hostname + connection-state dot (green / pulsing amber for copying|connecting / red for disconnected). Dot color is `f(connection.state)` — no parallel status enum.
- **`+ add host`** — inline form, invokes `hosts.add` admin RPC, persists.
- **`×` per tab** — invokes `hosts.remove`, closes any open browser WSes for that host server-side so partysocket can't reconnect.
- **Persistence** — `$XDG_STATE_HOME/drishti/hosts.json` (override via `DRISHTI_HOSTS_FILE`). Insertion order preserved end-to-end (hostsStore → registry → admin collection → tab strip).

### Pre-implement structural review (Hickey + Lowy)

Talk-mode artifact (`talk-multi-host.html`) ran both lenses on the design sketch _before_ any code. Folded-in revisions visible in the artifact body:

| Folded in | Why |
| --- | --- |
| Dropped `processCount` on `SystemInfo` | Lowy: god-cell driven by UI display volatility. Hickey: would force either two writers to one cell or fragmentation. |
[...]

PR #7, opened 2026-05-28
Show $PWD of processes

### LINKED ISSUE #6: Show $PWD of processes
For e.g., each of these claude's run in a different project directory and it would be useful to see them.

Image

srid:
**The process table now disambiguates same-command rows by surfacing each process's cwd.** Closes #6. Born from "I have five `claude` processes — which one is in which project?" the answer is now visible at a glance: a dim `@ /path/to/cwd` suffix appears after the command in the COMMAND cell.

### How it flows

```
/proc/<pid>/cwd ─▶ readProcLinuxRaw ─▶ ProcessSchema ─▶ ProcessRow
   (symlink)        readlink+.catch     cwd: z.string()    <Show>@ {cwd}
```

The new field is a peer scalar on the existing `processes` collection — no new abstraction, no new boundary. Linux reads `/proc/<pid>/cwd` via `readlink` alongside the existing parallel `/proc/<pid>/{stat,status,cmdline}` reads. The `.catch(() => "")` resolves EACCES (other-user pids) and ENOENT (kernel threads) to empty so one missing symlink can't drop the whole row.

### Per-platform behaviour

| Platform | cwd source | Empty when |
|---|---|---|
| **linux** | `readlink(/proc/<pid>/cwd)` | other-user pid (EACCES), kernel thread (ENOENT) |
| **darwin** | none — blank | always (lsof-per-pid would fork once per row) |
| **stub** | `process.cwd()` | never |

The UI hides the suffix when cwd is empty (`<Show when={proc()?.cwd}>`), so darwin and other-user rows look exactly as they did before. The filter input also matches cwd substrings, and the placeholder advertises it.

### Refinements landed during review

| Finding | Lens | Fix |
|---|---|---|
| Truncation literal `200` repeated at three call sites | Hickey | Extracted `PROC_STRING_MAX = 200`; all per-process string fields share one cap |
| Comment named `Promise.allSettled` while code used inline `.then(p, () => "")` | Hickey + Lowy | Rewrote comment to describe the actual rejection handler |
| `<Show>` re-read `proc()?.cwd` twice per render | Police (elegance) | Use the callback form `{(cwd) => <span>… {cwd()}</span>}` |
| Two-arg `.then(p => p, () => "")` reads as "transform-and-also-handle" | Police (elegance) | `.catch(() => "")` |

### Try it locally

```sh
nix run github:srid/drishti/show-process-cwd
```

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-7`)._

### DIFF
--- a/.agency/do.md
+++ b/.agency/do.md
@@ -22,6 +22,51 @@ reachable. -->
 ## Documentation
 Keep `README.md` in sync with user-facing changes.
 
-<!-- Optional (add manually for the evidence step):
 ## PR evidence
--->
+
+When the change has visible UI impact, post a `## Evidence` PR comment with screenshots. Use judgment — server-only diffs (host wiring, RPC plumbing) sometimes ripple into rendering and sometimes don't.
+
+**Delegate to a subagent** (`Agent(subagent_type="general-purpose", model="sonnet")`) so the main context stays clear of MCP and screenshot noise. Brief it with: the dev-server URL, what scenarios to capture, a `/tmp/drishti-evidence-<slug>.png` filename, and the PR number. Have it return only the markdown body it should post — the calling `/do` posts it with `gh pr comment`.
+
+### Dev server
+
+Spawn a dedicated dev server on a **free random port** (the user may already have one on 7720). The `port` flag — or `PORT` env var — is the override; both flow through `cli` in `packages/app/src/server/main.ts` and default to 7720.
+
+```sh
+PORT=$(python3 -c 'import socket; s=socket.socket(); s.bind(("",0)); p=s.getsockname()[1]; s.close(); print(p)')
+PORT=$PORT just dev localhost &
+DEV_PID=$!
+trap 'kill $DEV_PID 2>/dev/null' EXIT
+# wait for the server to print "listening on :$PORT" before driving the browser
+```
+
+For "before" shots on a bug fix, run a second server from a `git worktree` on `master` on a different free port. Never stash the PR branch.
+
+### Capture, host, post
+
+The subagent drives `chrome-devtools` MCP — `new_page` at `http://localhost:$PORT/`, waits for the htop table to populate (the agent needs one poll tick to seed the snapshot, ~2s), reproduces the relevant state, `take_screenshot` to `/tmp/drishti-evidence-<slug>.png`.
+
+`gh pr comment` can't attach binaries, so upload to a long-lived `evidence-assets` GitHub release on this repo and embed the download URL inline:
+
+```sh
+gh release view evidence-assets >/dev/null 2>&1 || \
+  gh release create evidence-assets --prerelease \
+    --title "Evidence assets (auto-uploaded by /do)" --notes "Do not delete."
+gh release upload evidence-assets /tmp/drishti-evidence-<slug>.png --clobber
+```
+
+URL pattern: `https://github.com/srid/drishti/releases/download/evidence-assets/<filename>`. Use the single-quoted heredoc pattern (`<<'EOF'`) when posting so backticks and `$` survive unescaped.
+
+### Process-table scenarios
+
+The interesting surface for drishti is the live process table — load average, memory, CPU strip, and the per-PID rows. Most PRs touching the UI want a capture of the table in its normal state, plus one for any new column / filter / dialog the diff introduces.
+
+Common scenarios worth a frame:
+
+- **Default load** — the table populated with ~hundreds of rows; proves the agent → parent → browser pipeline is alive end-to-end.
+- **Filtered view** — type into the filter input; capture the "showing N of M" counter shrinking.
+- **Sort change** — click a sortable header (PID / USER / CPU% / MEM%) and capture the descending indicator.
+- **Connecting overlay** — for changes that touch the `connection` cell (copying / connecting / disconnected states), spawn against an unreachable remote and capture the overlay.
+- **Multi-host tabs** — for changes that touch the admin surface or tab strip, launch with multiple hosts (`just dev localhost a.lan`) and capture the tab strip + connection dots.
+
+For PRs that add a per-process field (cwd, threads, fds, …), the default evidence is a **screenshot of the process table showing the new field for representative real processes** — that single frame proves the field flows schema → agent → wire → renderer end-to-end.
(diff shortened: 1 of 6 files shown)


srid:

| # | Lens   | Finding                                                          | Disposition         |
|---|--------|------------------------------------------------------------------|---------------------|
| 1 | Hickey | Truncation literal `200` fragmented across 3 call sites          | Fixed in this PR    |
| 2 | Hickey | Comment names `Promise.allSettled`; code uses inline `.then`     | Fixed in this PR    |
| 3 | Lowy   | Misleading `allSettled` comment in `readProcLinuxRaw`            | Fixed in this PR (same as #2) |
| 4 | Lowy   | `cwd: ""` does double duty (per-process failure vs darwin absence) | ⚠️ **No-op**       |

### Hickey rationale

The diff is mostly clean — additive, no new abstractions, all five concerns (acquisition, schema, dirty-check, presentation, docs) map to one structural layer each. No complecting, no concept multiplication. Two fragmentation findings:

**F1 — Truncation literal `200` fragmented across 3 sites.** `packages/app/src/agent/proc.ts` had three naked `truncate(..., 200)` call sites. The invariant "all per-process string fields share the same wire cap" was enforced only by convention. A change to one field's limit would silently break parity with the others. **Fix:** introduced `const PROC_STRING_MAX = 200` near the `truncate` helper and replaced all three call sites.

**F2 — Comment names `Promise.allSettled`, code uses `Promise.all`.** Around line 212 the comment stated the cwd readlink is "resolve[d] via `Promise.allSettled` so a failure here doesn't drop the whole row." The actual code was `Promise.all([..., readlink(...).then(p => p, () => "")])` — an inline rejection handler, not `allSettled`. The two APIs have different return shapes; a reader trusting the comment and refactoring toward literal `Promise.allSettled` would silently change behavior. **Fix:** rewrote the comment to accurately describe the inline handler.

**No-op:**
- `cwd: z.string()` as a required field with `""` sentinel — schema-documented, consistently produced by all three platform readers, and `<Show when={proc()?.cwd}>` correctly hides on the falsy empty string.
- `prev.cwd !== value.cwd` in `main.ts` dirty-check sits alongside the existing scalar comparisons.
- Inline `@ <cwd>` rendering inside the existing `COMMAND` `<td>` — no new component, no new state, filter extension reuses the same predicate structure as `user`/`command`.

### Lowy rationale

The decomposition is correctly placed. `cwd` lives inside existing volatility boundaries — *what process attributes are observable* (schema), *how to read procfs* (Linux reader), *how to read via `ps`* (darwin reader). No new module split is warranted, no duplicated receptacle, no inverted layering. The `prev.cwd !== value.cwd` predicate and inline `@ <path>` display follow the established per-field pattern.

**Note (variable, not volatile — no action):** `cwd: ""` does double duty — per-process failure (EACCES, kernel thread on Linux) **and** per-platform structural absence (darwin). The current UI hides both identically via `<Show when={proc()?.cwd}>`, and the README documents this. No current or roadmap consumer needs to distinguish the two states, so this is variability without volatility. Recording for future reference; not a fix.

**Fix in this PR — misleading `allSettled` comment** (same as Hickey F2 — both lenses converged).

### Cross-validation

Both reviewers cross-validated each other's findings. Neither lens identified a new problem introduced by the other's recommendations: `PROC_STRING_MAX` doesn't create a new volatility seam (it names an axis inside the existing receptacle), and the comment fix is pure documentation. Both passes endorsed each other's actions and the Lowy no-op on the empty-string sentinel.

srid:
## Evidence

The process table now shows each process's cwd as a dim `@ /path/to/cwd` suffix after the COMMAND text. Verified end-to-end on the dev server running against localhost (Linux `/proc/<pid>/cwd` reader → `ProcessSchema` → `ProcessRow`).

default state

Filtering on a cwd substring narrows the table to matching rows — here filtering by `drishti` shows only processes whose cwd resolves into the drishti worktree directories:

filtered by cwd

srid:

| research | ✓ | 1m 35s | Understood multi-host surface, Linux /proc reader, Darwin ps reader, client process table |
| branch | ✓ | 14s | Created `show-process-cwd` from `origin/master` |
| implement | ✓ | 1m 26s | Added cwd to ProcessSchema; Linux readlink; darwin/stub blank/process.cwd(); UI suffix + filter |
| check | ✓ | 36s | `bun run typecheck` clean |
| docs | ✓ | 11s | README processes table updated |
| fmt | ✓ | 8s | `just fmt` clean (0/10 .nix files) |
| commit | ✓ | 15s | `feat: show $PWD of processes` → e0ba471 |
| hickey+lowy | ✓ | 5m 40s | Two findings (PROC_STRING_MAX, allSettled comment) → commits 9cd254b, 452e73b; cross-validation clean |
| police | ✓ | **11m 21s** | Show callback form → 0cc9240; .catch over two-arg .then → 9769e1b; two findings dispositioned No-op |
| test | ✓ | 26s | `bun test` 7/7 pass |
| create-pr | ✓ | 1m 18s | Draft PR #7 + hickey/lowy comment |
| ci | ✓ | 1m 43s | justci all 14 nodes succeeded (x86_64-linux + aarch64-darwin) |
| evidence | ✓ | 3m 38s | Default + filtered screenshots via chrome-devtools MCP; re-CI on `.agency/do.md` HEAD also green |
| **Total** | | **28m 41s** | |

**Slowest step**: `police` (11m 21s)

### Optimization suggestions

- **`police` dominated the run (11m 21s, 40% of total).** Two reviewer invocations plus two follow-up fix commits — the first pass surfaced one Show double-accessor finding and ran ~6m; the re-invocation found one elegance finding and applied it itself. Worth seeing whether `/code-police` could batch the rule + fact-check + elegance passes into fewer round-trips when the diff is small (~6 files, +32/-7).
- **`hickey+lowy` was the second-largest at 5m 40s** (parallel reviewers + parallel cross-validators). The parallel-invocation discipline is already as good as it gets here; this is just inherent to running four sub-agents on the same diff.
- For follow-up runs on this branch, `--from ci-only` skips the 11m+ structural review when iterating on small fixes.
- Adding a `## PR evidence` section to `.agency/do.md` mid-run cost one extra CI re-run (1m 43s); for the next PR the section is already in place so this won't recur.

Workflow completed at 2026-05-28.

PR #1, opened 2026-05-27
Import remote-process-monitor as drishti

srid:
**drishti is a three-tier live process monitor** — browser SolidJS UI ↔ local Bun parent server ↔ remote stdio agent over `ssh`. The TS source is adapted from [juspay/kolu's `remote-process-monitor` example](https://github.com/juspay/kolu/tree/master/packages/surface/example/remote-process-monitor), pinned via npins, with kolu-plan commentary stripped and every `KOLU_*` identifier renamed to `DRISHTI_*`. The Nix scaffolding mirrors [srid/anywhen](https://github.com/srid/anywhen) verbatim — single `juspay/bun2nix` flake input, npins for nixpkgs+kolu, hand-rolled Bun.build pipeline. The repo itself was renamed `sshcient → drishti` mid-implementation via `gh repo rename`.

### Architecture

```
Browser (SolidJS UI)
   │  WebSocket (oRPC)
   ▼
Parent server (Bun, drishti)
   │  ssh stdio (oRPC)
   ▼
Remote host: drishti-agent
   │  /proc on linux, sysctl on darwin
   ▼
Kernel
```

The agent's `.drv` is resolved per target-host architecture (`ssh $host uname -ms`) so the binary built on the remote always matches its kernel. First connect ships the closure via `nix copy --derivation` then realises it remotely; subsequent connects reuse it.

### Nix scaffolding (anywhen-aligned)

- **One flake input**: `github:juspay/bun2nix/rawflake`. `nixpkgs` and `kolu` live in `npins/sources.json` — cold `nix develop` eval stays sub-second.
- **Three deployable derivations** behind one `drishtiBuilt`: `drishti-agent` (shipped via ssh), `drishti-client` (Bun.build bundle), `drishti` (`meta.mainProgram`, the `nix run` default).
- **Two `@kolu/*` source paths** materialised through a `mkKoluPackage` factory at `nix/packages/kolu-package.nix`, hydrated into `node_modules/@kolu/*` by `scripts/hydrate-kolu-packages.sh` (one script, three callers: shellHook, justfile, `postBunNodeModulesInstallPhase`).
- **Per-system outputs** via `eachSystem` over `[x86_64-linux, aarch64-linux, aarch64-darwin]`.
[...]

PR #10, opened 2026-05-29
Add per-interface network I/O metrics

srid:
**drishti showed CPU, memory, load, command, and cwd — but zero network.** This adds a live per-NIC view: receive/transmit throughput (bytes/sec) plus cumulative totals, for every interface, flowing through the same reactive transport as the rest of the surface. It closes the most-validated gap in the feature plan (rtop, drishti's closest agentless peer, defines its value around exactly this).

### How it fits together

A new `networkInterfaces` collection — keyed by NIC name — rides the existing agent → parent → browser pipeline, modeled cell-for-cell on the `cpuCores` collection that was already there:

```
/proc/net/dev (linux) ─┐
netstat -ib  (darwin) ─┼─▶ ProcReader.readNetwork ─▶ networkInterfaces ─▶ parent mirror ─▶ NetStrip
(unknown OS → empty)  ─┘   (rate-from-delta closure)   Collection<name,…>   pumpNetwork…     (one rx/tx row per NIC)
```

Throughput is a *rate*, so the agent retains the previous tick's cumulative counters and derives bytes/sec from the delta — the same shape the per-core CPU reader uses. The parsing and the rate math are pure, exported functions (`parseProcNetDev`, `parseNetstatIb`, `computeNetThroughput`), so the part that's wrong at runtime if a formula slips is unit-tested directly.

### Notable decisions

- **Loopback is filtered at the reader, not the parser.** The parsers stay faithful to their source (a test asserts `lo` is parsed); "what the app wants to show" is a separate policy applied after. `lo`/`lo0` carry intra-host traffic, not network I/O, and would otherwise dominate the list.
- **Counters that run backwards clamp to 0** — a NIC reset or hot-swap reports no throughput rather than a negative or absurd spike.
- **`netstat -ib` columns are counted from the right**, because the optional MAC-address column shifts absolute field positions; the byte counters are stable relative to the end of the line.

### Refinements during structural review

The Hickey/Lowy + code-police passes landed five follow-up commits. The load-bearing one was a real bug:

| What was off | Fix |
| --- | --- |
| `formatBytes(999_999)` rendered `"1000.0 KB"` — the unit was chosen *before* rounding to one decimal | Promote one more unit when rounding hits the boundary; added a regression test |
| `CpuStrip`/`NetStrip` duplicated byte-identical strip chrome | Extracted a generic `MetricStrip<T>`; both are now thin wrappers |
| `createNetReader` held the previous counters and their timestamp as two separate `let`s | Collapsed into one record so the per-tick swap is atomic |
| `FragmentCtx` was documented as a "keep in sync" chore | Reframed as a deliberate least-privilege narrowing with a compile-time drift guard |

### Try it locally

```sh
nix run github:srid/drishti/feat/network-io
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Host 1: drishti-agent     Host 2: drishti-agent     …
 Kernel
 ```
 
-Host identity lives only at the transport layer: each browser tab opens its own WebSocket to `/rpc/ws?host=<id>`; the parent dispatches to a per-host `RPCHandler` (built once per host from the same surface schema). The per-host `surface` schema (system / processes / cpuCores / connection cells) is scalar — no host dimension anywhere in the primitives.
+Host identity lives only at the transport layer: each browser tab opens its own WebSocket to `/rpc/ws?host=<id>`; the parent dispatches to a per-host `RPCHandler` (built once per host from the same surface schema). The per-host `surface` schema (system / processes / cpuCores / networkInterfaces / connection cells) is scalar — no host dimension anywhere in the primitives.
 
 A separate **admin surface** at `/rpc/ws?host=__admin__` exposes the *set* of hosts as a `Collection<string, HostEntry>` plus `addHost` / `removeHost` procedures. The tab strip subscribes to the collection; the `+` / `×` buttons call the procedures.
 
@@ -50,6 +50,7 @@ Per-host primitives:
 | **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, memPct, command, cwd }`. Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
 | **Stream** | `processesSnapshot` | Bulk-snapshot variant for ~600-PID htop refresh in one frame. |
 | **Collection** | `cpuCores` | Per-core CPU usage (`Collection<K,T>` showcase). |
+| **Collection** | `networkInterfaces` | Per-NIC network I/O, keyed by interface name — `{ rxBytes, txBytes, rxRate, txRate }` (cumulative bytes since boot + bytes/sec throughput). `/proc/net/dev` on linux, `netstat -ib` on darwin; loopback filtered out. |
 | **Procedure** | `process.kill` | The only mutation — sends `TERM` / `KILL` / `HUP` / `INT`. |
 
 Admin surface primitives:
--- a/packages/app/src/agent/main.ts
+++ b/packages/app/src/agent/main.ts
@@ -37,6 +37,8 @@ import {
   type CoreId,
   type CpuCore,
   DEFAULT_CONNECTION,
+  type IfaceName,
+  type NetInterface,
   type Pid,
   type Process,
   type ProcessesSnapshotMsg,
@@ -85,6 +87,11 @@ async function main(): Promise<void> {
   const cpuCoreSnapshot = new Map<CoreId, CpuCore>();
   for (const [core, value] of reader.readCpuCores())
     cpuCoreSnapshot.set(core, value);
+  // Seed the network baseline so the first published delta has a previous
+  // tick to rate against — this first read returns 0 bytes/sec everywhere.
+  const netSnapshot = new Map<IfaceName, NetInterface>();
+  for (const [iface, value] of await reader.readNetwork())
+    netSnapshot.set(iface, value);
 
   // Build the surface implementation. The `processes` collection's
   // `readAll` yields the current snapshot; `upsert`/`remove` are the
@@ -125,6 +132,15 @@ async function main(): Promise<void> {
         },
         remove: (key) => {
           cpuCoreSnapshot.delete(key);
+        },
+      },
+      networkInterfaces: {
+        readAll: () => netSnapshot,
+        upsert: (key, value) => {
+          netSnapshot.set(key, value);
+        },
+        remove: (key) => {
+          netSnapshot.delete(key);
         },
       },
     },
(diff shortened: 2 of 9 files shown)


srid:

Both lenses ran as sub-agents on the feature commit, then a cross-validation pass ran in both directions (each lens auditing the other's recommendations). Every finding landed as its own follow-up commit — **none deferred**.

| 1 | Hickey | `createNetReader` holds previous counters + timestamp as two `let`s for one concept | Fixed in this PR |
| 2 | Hickey | `FragmentCtx` documented as a "keep in sync" chore rather than an intentional narrowing | Fixed in this PR |
| 3 | Hickey (cross-val) | `formatBytes` selects the unit *before* `.toFixed(1)` → `999_999` renders `"1000.0 KB"` | Fixed in this PR |
| 4 | Hickey (cross-val) | `CpuStrip`/`NetStrip` duplicate byte-identical strip chrome | Fixed in this PR |
| 5 | Hickey (cross-val) | `rxTotal`/`txTotal` single-use signals add indirection with no reactive benefit | Fixed in this PR |
| 6 | Lowy (cross-val) | Confirms #1/#2 safe under the volatility lens; #2 should be framed as least-privilege | Fixed in this PR (folded into #2) |

> Lowy's first pass returned **no problem findings** — it confirmed the diff *extends* the canonical in-repo patterns (`cpuCores` collection, `ProcReader`, `createCpuCoresReader`, `mirrorRemoteCollection`, `CpuStrip`, `metrics.ts` formatters) for every operation kind rather than reinventing them, and that the three change axes (per-OS counter source / throughput derivation / presentation) each sit behind their own boundary.

### Hickey rationale

The duplication audit came back clean at the headline level — `NetStrip`/`NetCell`, `createNetReader`, and `formatBytes` are parallel instantiations of domain-natural patterns, not redundant second concepts. The actionable findings were smaller: a fragmented snapshot (`prev`/`prevMs`) that should advance atomically, and the `FragmentCtx` hand-narrowing whose comment framed it as a maintenance burden. The cross-validation pass — running the Hickey lens against the code Lowy had endorsed as sound — surfaced the load-bearing one: a **real correctness bug** in `formatBytes` rounding at unit boundaries, plus the `CpuStrip`/`NetStrip` chrome duplication (extracted into a generic `MetricStrip<T>`).

### Lowy rationale

The decomposition encapsulates the three axes actually likely to change: the per-OS counter source behind `ProcReader.readNetwork()` (swapping `/proc/net/dev` touches only `linuxReader`), throughput derivation behind the pure `computeNetThroughput` (clock isolated in the closure so it stays testable), and presentation behind `formatBytes`/`NetStrip`. `filterLoopback` is correctly placed in the *reader*, not the *parser* — "what the file contains" and "what the app wants" are separate volatility axes. Cross-validating Hickey's two recommendations found both safe; #2 (`FragmentCtx`) was additionally a net positive — it removes a cross-module replica of the surface shape — so the fix reframes it as a deliberate least-privilege boundary with a compile-time drift guard.

srid:
## Evidence

Network I/O strip on localhost host view

Live per-NIC network throughput strip (labeled "network (5)") rendered directly below the CPU cores strip on the localhost host view, showing real `↓`/`↑` rates (e.g. `↓ 6.1 KB/s ↑ 14.4 KB/s` for `enp1s0`, `↓ 1.3 KB/s ↑ 14.0 KB/s` for `tailscale0`) after the second poll tick. Loopback is filtered out as designed. Captured end-to-end through the agent (`/proc/net/dev`) → parent mirror → browser pipeline.

srid:

|------|--------|----------|-------------|
| research | ✓ | 3m 38s | Mapped cpuCores collection analog + agent rate-from-delta pattern; identified /proc/net/dev + netstat -ib sources |
| branch | ✓ | 0s | feat/network-io from origin/master |
| implement | ✓ | 3m 34s | Surface collection + agent readers + loop publish + client NetStrip + formatters; tests added |
| check | ✓ | 1m 17s | tsc --noEmit clean (incl. parent router mirror wiring) |
| docs | ✓ | 26s | README primitives table + surface-schema sentence updated |
| fmt | ✓ | 14s | nixpkgs-fmt clean; no JS formatter configured |
| commit | ✓ | 18s | e66af20 pushed |
| **hickey+lowy** | ✓ | **18m 45s** | Both lenses + bidirectional cross-validation; 6 findings, all Fixed (0 deferred); 5 follow-up commits |
| police | ✓ | 3m 16s | rules + fact-check + elegance all clean |
| test | ✓ | 11s | 21 pass / 0 fail (parsers, rate math, formatters incl. boundary regression) |
| create-pr | ✓ | 1m 8s | Draft PR #10 + Hickey/Lowy ledger comment |
| ci | ✓ | 1m 48s | justci: all 14 nodes green (x86_64-linux + aarch64-darwin) on HEAD |
| evidence | ✓ | 2m 57s | ## Evidence comment with verified network-strip screenshot |
| **Total** | | **37m 54s** | |

### Optimization suggestions

- **`hickey+lowy` dominated (18m 45s, ~49% of wall-clock).** The cost was four serial sub-agent rounds (two first-pass + two cross-validation). Lowy's first pass returned only No-ops, yet its cross-validation half still ran — for a diff that cleanly *extends* an established pattern (here, `cpuCores`), the zero-actionable-findings case is a candidate to short-circuit the reverse cross-validation direction.
- **One cross-validation sub-agent burned time reconstructing the wrong findings before re-deriving them from the quoted diff.** Pasting the *exact* prior-pass findings verbatim (which this run did) is right; the slowness was internal to the skill. Worth noting the cross-validation caught a real `formatBytes` rounding bug Lowy's lens couldn't see — the time bought correctness.
- **`research` (3m 38s)** could shrink by pre-reading the closest in-repo analog before invoking — the whole feature mirrors `cpuCores` end-to-end (`surface.ts`, `proc.ts`, `main.ts`, `router.ts`, `App.tsx`). Skimming those five files first would cut the exploration.

Workflow completed at 2026-05-29 (HEAD 15f9fa8).

PR #8, opened 2026-05-29
Add a fleet overview — every host's health on one pane

srid:
**drishti now opens on a fleet overview: one live summary card per configured host — connection state, CPU (mean across cores), memory, load average, uptime, OS — in a single pane.** Until now the only way to read N hosts was to click through N tabs; the aggregate "single pane of glass" is the most-validated value proposition for multi-host monitors, and drishti already had every piece needed to build it. Clicking a card (or a host tab) drills into that host's full htop body.

The whole feature is **pure client-side aggregation** — no schema, agent, or server changes. Each card subscribes to the host's existing `system`/`connection` cells and `cpuCores` collection over the *same per-host socket the tab chips already keep warm*, so the overview costs subscriptions, not new connections.

```
fleet tab ─▶ FleetView ─▶ HostCard × N
                              │  surfaceForHost(h)  (cached socket)
                              ▼
                  system · connection · cpuCores   ──reactive──▶ live card
```

### What you see

- **fleet tab** (leftmost, default landing view) — a responsive grid of host cards; a per-host connection dot pulses while provisioning/connecting
- **per-card metrics** — CPU and memory usage bars, load average, uptime, OS, core count
- **not-connected cards** — show a terse status (`provisioning agent…` / `connecting…` / `no data`) instead of stale zeros
- **click to drill in** — a card or a host tab switches to that host's existing process table; removing the active host falls back to the fleet view rather than a blank pane

### Structure

The view selection is modelled as a sum — `View = {kind:"fleet"} | {kind:"host"; host}` (in its own `view.ts`) — so "which host" only exists in the branch where a host is selected. Two new Solid-free, unit-tested helper modules carry the reused logic:

| Module | Owns |
| --- | --- |
| `metrics.ts` | `memPct` / `memGb` / `formatUptime` / `averageCoreUsage` — derivations the per-host header and the fleet cards share |
| `usageColors.ts` | the emerald/amber/red usage palette, single-sourced across the memory bar, per-core bars, and process columns |
| `connectionColors.ts` | one `STATE` map: dot colour, text colour, terse label, verbose message, and pending flag per `ConnectionState` |

> _Structural review (hickey + lowy, with a cross-validation pass) drove the helper extractions and the `View` placement — see the analysis comment below._

### Try it locally

```sh
nix run github:srid/drishti/feat/fleet-dashboard -- localhost a.lan b.lan
```

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

PR #5, opened 2026-05-28
Cross-arch agent provisioning: macOS → Linux without Exec format error

srid:
**A macOS user running `drishti user@linux-host` no longer ships the darwin closure across `nix copy --derivation`.** The monitor wrapper now bakes a `{system → drvPath}` map at build time, and the parent server picks the matching `.drv` per host via a `uname -ms` probe on add — using [`@kolu/surface-nix-host`'s `resolveSystem`](https://github.com/juspay/kolu/pull/1009). One `nix run` invocation can mix architectures freely.

### What broke before

The wrapper used to bake a single `DRISHTI_AGENT_DRV` — whatever the build evaluator's system was. On macOS, that was an aarch64-darwin derivation. Adding a Linux host shipped that .drv across; the Linux daemon substituted the prebuilt darwin closure from cache.nixos.org; exec'ing the wrapper died with:

> `/nix/store/…/bin/drishti-agent: line 2: /nix/store/…-bun-1.3.10/bin/bun: cannot execute binary file: Exec format error`

The kolu library was already arch-agnostic on purpose ([`nixCopy.ts:7-13`](https://github.com/juspay/kolu/blob/master/packages/surface-nix-host/src/nixCopy.ts#L7-L13)) — _the caller is the right place to pick the .drv_. Drishti just wasn't doing it.

### How the new flow fits together

```
flake.nix                  default.nix              packages/app/src/server
─────────                  ───────────              ──────────────────────
systems = […]              monitor wrapper          on add(host):
   │                          bakes                    1. resolveSystem(host)
   ▼                       DRISHTI_AGENT_DRVS_JSON       (kolu helper — uname -ms,
agentDrvBySystem  ───────▶ '{…}'              ───▶       locally or via ssh)
   = { sys → drvPath                                  2. resolveDrvForHost(host, map)
       per system }                                   3. getHostSession({ host, drvPath })
```

`agentDrvBySystem` is **pure eval** — `drvPath` is just a string, no IFD — so the macOS evaluator can produce the linux drv path (and vice versa) without remote builders. `builtins.unsafeDiscardStringContext` keeps the cross-system drv paths from pulling their full closures into the wrapper's realised output.

### Companion kolu PR

[...]

PR #4, opened 2026-05-28
ci: work around crates.io 403 and unreachable linux runner

srid:
**`nix run github:juspay/justci` was failing on both lanes** before this branch — neither problem is in our code, but both have repo-local workarounds that make CI green again without waiting on upstream.

### Failure modes

| Lane | Symptom | Root cause |
|------|---------|-----------|
| `ci::nix@aarch64-darwin` | `curl: (22) The requested URL returned error: 403` on every `crate-*.tar.gz` fetch | crates.io's anti-bot now rejects `curl/*` User-Agent; nixpkgs `fetchurl` uses curl's default UA. `bun2nix-2.1.0`'s rust dep tree (askama, basic-toml, jsonc-parser, …) is unbuildable. |
| `_ci-setup@x86_64-linux` | `ssh -T srid-justci` exits 255 with `ERR not owner of srid-justci` | The Tailscale proxy alias for the configured linux runner refuses our identity. The whole linux pipeline is blocked at setup. |

### Fixes in this PR

- **`scripts/ci-prefetch-crates.sh` + `_prefetch-crates` recipe.** New `ci/mod.just` node, gated as a dep of `nix:`. The script walks the bun2nix build closure, finds `crate-*.tar.gz` FODs whose output isn't valid in the store yet, fetches each with `curl -A Mozilla/5.0`, and injects it via `nix-store --add-fixed sha256`. Idempotent on warm stores (no network), runs once per platform, content-addressed so the artifact survives any future `bun2nix` bump that doesn't change the crate hashes.
- **`.agency/do.md` CI command** now passes `--host x86_64-linux=localhost`. `/do` runs the linux lane on the invoking machine until `srid-justci` is reachable again.

### Permanent fixes proposed

These should be filed upstream — once any of them lands, the matching workaround in this PR becomes deletable code:

| Problem | Permanent fix | Owner |
|---------|--------------|-------|
| crates.io 403 | nixpkgs `fetchurl` should set a non-`curl/*` UA by default (or special-case `crates.io/api/v1/...` to hit `static.crates.io` directly) | nixpkgs |
| crates.io 403 (alt) | `bun2nix`'s rust crate fetcher should target `static.crates.io/crates/<name>/<name>-<ver>.crate` directly, bypassing the bot-filtered API | `juspay/bun2nix` |
| `srid-justci` unreachable | Restore the Tailscale proxy, or repoint `~/.config/justci/hosts.json` to a different x86_64-linux runner | infra |

[...]

PR #11, opened 2026-05-29
Per-host time-series charts with switchable history windows

srid:
**Each host's htop view now carries a live time-series chart** — CPU% and memory% plotted over a rolling history, with a segmented control to switch the visible window (1m / 5m / 15m / 30m) the way btop and Netdata do. drishti previously showed only live state; this adds the "what happened 30 seconds ago" view without giving up the zero-config posture.

The history is a **server-side in-memory ring, owned by the parent**. The parent observes every agent poll tick regardless of which browser tabs are open, so it samples CPU%/mem% per host into a ring that lives for the life of the process. The browser is a pure reader: it subscribes to a new `metricHistory` stream and is re-seeded with the entire ring on every connect.

> **Persistence:** because the ring lives in the parent, the chart **survives page reloads and tab switches**, and is already populated the first time you open a host. It's in-memory only — restarting the parent starts fresh (no disk, preserving zero-config). A persistent/exporting store (Prometheus/InfluxDB) remains the deliberately-deferred larger step from the plan.

### How it fits together

```
agent poll tick ─▶ parent: captureSample() ─▶ pushSample() ─▶ historyRing (per host)
                                                                   │
                                              metricHistory stream (snapshot-then-delta)
                                                                   ▼
                              browser: subscribe ─▶ history signal ─▶ HistoryChart (pure SVG)
```

`metricHistory` mirrors the existing `processesSnapshot` shape: a fresh subscriber gets the whole ring (`snapshot`), then one `sample` per tick (`delta`). The agent serves an inert empty stub — it keeps no history; the parent is authoritative. The pure history domain (ring eviction, windowing, SVG projection, sample capture) lives in `common/history.ts`, shared by both tiers; numeric derivations (`memPct`, `averageCoreUsage`) live in `common/metrics.ts`.

### What you get

- **Dual sparkline** — CPU (emerald) and memory (indigo) overlaid, filling from the right as data accrues
- **Window switcher** — 1m / 5m / 15m / 30m, instant against the buffered samples
- **Live legend** — latest CPU% / mem% beside the trace
- **Survives reload & tab switch** — server-owned ring, replayed on connect

### Structural refinements during review

Two rounds of hickey/lowy + a code-police pass followed the feature commits (see the analysis comments). Highlights:

| What a lens caught | Fix |
| --- | --- |
| History died on tab switch / reload (client-owned) | **Ring moved to the parent server** (this is the headline of the second round) |
| `HistoryChart` braided windowing math with SVG rendering | Pure renderer; projection lifted into `HostView` |
| Section chrome duplicated across three components | Extracted a single `MetricSection` |
| `Sample` construction scattered | `captureSample` factory in `common/history.ts` |
| Two `AbortController`s with one lifecycle; inert agent channel never published | Collapsed to one controller; parked the stub on the abort signal |

> A notable cross-validation catch: Hickey proposed the client drop its `pushSample` and plain-append deltas; Lowy's cross-check showed that would grow the client array unboundedly over a long session (the delta stream is unbounded — only the *snapshot* is pre-bounded), so the client-side bound stayed.

### Try it locally

```sh
nix run github:srid/drishti/hisstory
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,6 @@ node_modules/
 dist/
 state/
 
-# Session artifacts (talk-mode HTML) — local design scratch, not for the PR
-/talk-*.html
-
 # Nix
 /result
 /result-*
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ nix run github:srid/drishti -- localhost a.lan b.lan   # multiple hosts (tabbed 
 
 Open <http://localhost:7720>. The UI opens on the **fleet** tab — a single overview pane with one live summary card per host (connection state, CPU, memory, load average, uptime); click a card (or a host's tab) to drill into that host's full htop. Each host also has its own tab with a live connection-state dot.
 
+A host's htop view carries a **time-series chart** above the CPU strip: CPU% and memory% plotted over a rolling, in-memory history sampled at the host's poll cadence. A segmented control switches the visible window (1m / 5m / 15m / 30m), like the time-range chips in btop and Netdata. The history is ephemeral — held in the browser for the life of the tab, never persisted — so it costs nothing to keep and starts fresh on reload.
+
 Every view has its own URL, so you can bookmark or share a link straight to a host: selecting a host updates the address to <http://localhost:7720/?host=user@host> (the fleet overview is the bare <http://localhost:7720/>). Opening such a link, or reloading the page, lands directly on that host.
 
 Use the `+ add host` button in the tab strip to add hosts at runtime; the `×` on each tab removes one. Added/removed hosts persist to `$XDG_STATE_HOME/drishti/hosts.json` (override with `DRISHTI_HOSTS_FILE`), so `nix run github:srid/drishti` with no args restores the last session.
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Per-host primitives:
 | Primitive | Path | Purpose |
 |---|---|---|
 | **Cell** | `system` | Load averages, memory, uptime, OS, hostname. |
-| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, memPct, command, cwd }`. Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
+| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, rssBytes, command, cwd }`. The table shows absolute resident memory (`rssBytes`, auto-scaled to MB/GB). Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
 | **Stream** | `processesSnapshot` | Bulk-snapshot variant for ~600-PID htop refresh in one frame. |
 | **Collection** | `cpuCores` | Per-core CPU usage (`Collection<K,T>` showcase). |
 | **Collection** | `networkInterfaces` | Per-NIC network I/O, keyed by interface name — `{ rxBytes, txBytes, rxRate, txRate }` (cumulative bytes since boot + bytes/sec throughput). `/proc/net/dev` on linux, `netstat -ib` on darwin; loopback filtered out. |
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ nix run github:srid/drishti -- localhost a.lan b.lan   # multiple hosts (tabbed 
 
 Open <http://localhost:7720>. The UI opens on the **fleet** tab — a single overview pane with one live summary card per host (connection state, CPU, memory, load average, uptime); click a card (or a host's tab) to drill into that host's full htop. Each host also has its own tab with a live connection-state dot.
 
-A host's htop view carries a **time-series chart** above the CPU strip: CPU% and memory% plotted over a rolling, in-memory history sampled at the host's poll cadence. A segmented control switches the visible window (1m / 5m / 15m / 30m), like the time-range chips in btop and Netdata. The history is ephemeral — held in the browser for the life of the tab, never persisted — so it costs nothing to keep and starts fresh on reload.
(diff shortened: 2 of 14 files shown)


srid:

Both lenses ran as parallel sub-agents on the primary feature commit, then a cross-validation pass had each lens audit the other's recommendations. Findings ledger:

| 1 | Hickey | `HistoryChart` takes both `windowKey` and a pre-resolved `windowMs` (one selection, two props) | Fixed in this PR |
| 2 | Hickey | Section chrome (`border-b … px-4 py-2`) duplicated across `MetricStrip` / `FilterBar` / `HistoryChart` | Fixed in this PR |
| 3 | Lowy | `Sample` construction split between its type (history.ts) and the inline capture site (HostView) | Fixed in this PR |
| 4 | Lowy (cross-val) | Hickey #1, applied naively, deepens the projection-vs-rendering braid in `HistoryChart` | Fixed in this PR (made it a pure renderer; projection moved up to `HostView`) |
| 5 | Hickey | Inline / genericize `DurationPicker` | ⚠️ **No-op** |
| 6 | Lowy | Decouple `HISTORY_RETENTION_MS` from `Math.max(...windows)` into an independent constant | ⚠️ **No-op** |
| 7 | Lowy | Make `windowSlice` / `polylinePoints` private (single consumer) | ⚠️ **No-op** |
| 8 | Lowy | Move chart series colors into `usageColors.ts` | ⚠️ **No-op** |

### Hickey rationale

The file split is structurally sound — pure ring/windowing/projection math in `history.ts` parallels the existing `metrics.ts` pure-derivation pattern; the diff extends it rather than reinventing it. Fixable findings were all in the new UI components: a redundant prop on `HistoryChart` (#1) and the section-chrome fragmentation (#2), both fixed.

Cross-validating Lowy's recommendations, the Hickey lens **rejected three**: decoupling `HISTORY_RETENTION_MS` (#6) would reintroduce a magic value and downgrade an invariant *enforced by construction* to one *checked by a test* — the computed `Math.max(...)` form guarantees the ring always retains the widest selectable window; privatizing `windowSlice`/`polylinePoints` (#7) would hide the highest-risk math (coordinate inversion, clamping, time-scaling) from its unit tests, contradicting the module's stated purpose and the `metrics.ts` precedent; and folding the chart's *categorical* series colors into the *severity-ramp* `usageColors.ts` (#8) would complect two unrelated concepts — the codebase already keeps `NetCell`'s categorical rx/tx colors inline for exactly this reason.

### Lowy rationale

`history.ts` cleanly encapsulates the metric-history domain. The actionable finding was that `Sample` assembly was fragmented between the type and the inline capture expression in `HostView` — a `captureSample` factory (#3) gives the concept one home (reusing `averageCoreUsage`/`memPct` rather than re-deriving them, per Hickey cross-validation).

Cross-validating Hickey's recommendations, the Lowy lens **refined #1 and rejected #5**: Hickey #1 was right that `windowMs` shouldn't be a redundant prop, but resolving it *inside* `HistoryChart` would deepen the projection-vs-rendering braid — so the projection memos moved *up* into `HostView` and `HistoryChart` became a pure renderer (#4). `DurationPicker` (#5) is the canonical single-axis receptacle for window selection; inlining it would fuse control-presentation into chart-rendering, and genericizing it would fragment the window-definition axis across the call site with no second consumer to justify the abstraction — so it stays as-is.

No deferred findings — every finding is either fixed in this PR or a No-op (the diff is already correct under that lens).

srid:
## Evidence

Dual CPU/memory sparkline chart at default 5m window
*The "history" panel in the htop view, showing populated dual SVG sparklines (emerald = CPU%, indigo = memory%) over the default 5-minute rolling window, with the segmented 1m / 5m / 15m / 30m duration control on the right and a live legend displaying the latest cpu/mem percentages.*

Same history chart after switching to the 1m window
*After clicking the "1m" chip: the window narrows to the most recent minute of samples, demonstrating the duration control switching the visible trace range in real time.*

srid:

| sync | ✓ | 1s | forge=github; noGit=false |
| research | ✓ | 5m 7s | Mapped pipeline: client-only feature; ring/window/projection factored to pure history.ts; chart mounts between Header and CpuStrip |
| branch | ✓ | 0s | Reused existing clean feature branch `hisstory` at origin/master HEAD |
| implement | ✓ | 0s | Pure history.ts + history.test.ts; HostView timer sampling + HistoryChart/DurationPicker |
| check | ✓ | 33s | tsc --noEmit exited 0 |
| docs | ✓ | 0s | README Quick start describes the chart + 1m/5m/15m/30m switcher |
| fmt | ✓ | 17s | nixpkgs-fmt clean (no JS formatter configured) |
| commit | ✓ | 7s | Primary feature commit b40b3c2 pushed |
| **hickey+lowy** | ✓ | **12m 47s** | Both lenses + cross-validation; 4 findings fixed (own commits), 4 No-op, no defers |
| police | ✓ | 7m 13s | 3 passes clear; invalid-states + redundant-label fixes committed individually |
| test | ✓ | 19s | bun test 55/55; history.test.ts exercises the new logic |
| create-pr | ✓ | 55s | Draft PR #11 + Hickey/Lowy analysis comment |
| ci | ✓ | 1m 54s | justci all 14 nodes succeeded (linux + darwin), against HEAD |
| evidence | ✓ | 2m 54s | Captured & posted populated-chart screenshots (5m + 1m windows) |
| **Total** | | **34m 16s** | |

**Slowest step**: `hickey+lowy` (12m 47s)

### Optimization suggestions

- **`hickey+lowy` dominated (12m 47s)** because it ran four sub-agents — two first-pass lenses plus the mandatory cross-validation pair. The cost was justified *here* (cross-validation flipped Hickey F2 to No-op and reshaped F1 into the pure-renderer refactor), but for a tightly-scoped diff that's confidently confined, `--minimal` skips both this and `police` (~20m combined) when structural review is overkill.
- **`research` took 5m 7s** — mostly an Explore fan-out over the metrics pipeline. Pre-reading `client/App.tsx` + `common/surface.ts` before invoking `/do` would let research start from a known map and skip the broad sweep.
- **`police` cost 7m 13s**, all in the elegance pass's forked sub-agents; it caught two real redundancies (`hasSamples`, `label`). No flakiness — no action needed beyond the `--minimal` note above for trivial diffs.
- For any post-merge CI re-runs, `--from ci-only` skips straight to the 1m 54s CI step.

Workflow completed 2026-05-29.

PR #12, opened 2026-05-29
Show absolute process memory; fix frozen header mem readout

srid:
The per-host detail header had a frozen memory readout. SolidJS runs a component body exactly once, at mount — and at that moment `system` is still `DEFAULT_SYSTEM` (`memUsed`/`memTotal` both 0). The `Header` derived `pct`/`gb` as plain consts from that initial value, so **the memory line stayed pinned at `0.0/0.0 GB (0%)` for the life of the view**, even as load, uptime, OS, and hostname updated (those are read inline in JSX, so they stayed reactive). The fleet `HostCard` already does this correctly with `createMemo`; this change mirrors it in the header and replaces the misleading comment that claimed the body re-runs on prop change.

Alongside the fix, the per-process **MEM%** column is replaced with absolute resident memory — **MEM** — auto-scaled to MB/GB (`1.2 GB`, `340 MB`) via the existing decimal `formatBytes` helper. A percent-of-total is rarely what you want when triaging a host; the absolute footprint is.

### What changed in the wire surface

The process schema's `memPct` field is **replaced** by `rssBytes` (resident set size, in bytes) — not duplicated. One domain concept, one field, consumed end-to-end:

| Layer | Before | After |
| --- | --- | --- |
| `ProcessSchema` (common) | `memPct: number` | `rssBytes: number` |
| linux reader | `100 × VmRSS·1024 / memTotal` | `VmRSS·1024` (kB → bytes) |
| darwin `ps` format | `…,pcpu=,pmem=,comm=` | `…,pcpu=,rss=,comm=` (rss is kB → bytes) |
| sort + delta diff | by `memPct` | by `rssBytes` |
| table cell | `mem.toFixed(1)` % | `formatBytes(rssBytes)` |

Keeping darwin and linux _in parity_ drove the design: both platforms now report exactly the absolute bytes the UI renders, and the dead `pmem=` column / linux percentage computation are gone rather than left to drift. The darwin parse path is extracted into a pure, exported `parsePsLine` so the format change sits behind one regex and one test.

## Testing
- `just typecheck` — clean
- `bun test packages/app` — 45 pass (3 new `parsePsLine` cases: rss→bytes, space-bearing commands, malformed lines)
- `just fmt-check` — clean

### Try it locally

```sh
nix run github:srid/drishti/mem
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Per-host primitives:
 | Primitive | Path | Purpose |
 |---|---|---|
 | **Cell** | `system` | Load averages, memory, uptime, OS, hostname. |
-| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, memPct, rssBytes, command, cwd }`. The table shows absolute resident memory (`rssBytes`, auto-scaled to MB/GB) rather than `memPct`. Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
+| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, rssBytes, command, cwd }`. The table shows absolute resident memory (`rssBytes`, auto-scaled to MB/GB). Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
 | **Stream** | `processesSnapshot` | Bulk-snapshot variant for ~600-PID htop refresh in one frame. |
 | **Collection** | `cpuCores` | Per-core CPU usage (`Collection<K,T>` showcase). |
 | **Collection** | `networkInterfaces` | Per-NIC network I/O, keyed by interface name — `{ rxBytes, txBytes, rxRate, txRate }` (cumulative bytes since boot + bytes/sec throughput). `/proc/net/dev` on linux, `netstat -ib` on darwin; loopback filtered out. |
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Per-host primitives:
 | Primitive | Path | Purpose |
 |---|---|---|
 | **Cell** | `system` | Load averages, memory, uptime, OS, hostname. |
-| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, memPct, command, cwd }`. Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
+| **Collection** | `processes` | Keyed by PID — `{ user, cpuPct, memPct, rssBytes, command, cwd }`. The table shows absolute resident memory (`rssBytes`, auto-scaled to MB/GB) rather than `memPct`. Snapshot-then-delta. `cwd` is from `/proc/<pid>/cwd` on linux (empty on darwin / kernel threads / other-user pids). |
 | **Stream** | `processesSnapshot` | Bulk-snapshot variant for ~600-PID htop refresh in one frame. |
 | **Collection** | `cpuCores` | Per-core CPU usage (`Collection<K,T>` showcase). |
 | **Collection** | `networkInterfaces` | Per-NIC network I/O, keyed by interface name — `{ rxBytes, txBytes, rxRate, txRate }` (cumulative bytes since boot + bytes/sec throughput). `/proc/net/dev` on linux, `netstat -ib` on darwin; loopback filtered out. |
--- a/packages/app/src/agent/main.ts
+++ b/packages/app/src/agent/main.ts
@@ -194,7 +194,6 @@ async function main(): Promise<void> {
         if (
           prev === undefined ||
           prev.cpuPct !== value.cpuPct ||
-          prev.memPct !== value.memPct ||
           prev.rssBytes !== value.rssBytes ||
           prev.command !== value.command ||
           prev.cwd !== value.cwd
--- a/packages/app/src/agent/main.ts
+++ b/packages/app/src/agent/main.ts
@@ -195,6 +195,7 @@ async function main(): Promise<void> {
           prev === undefined ||
           prev.cpuPct !== value.cpuPct ||
           prev.memPct !== value.memPct ||
+          prev.rssBytes !== value.rssBytes ||
           prev.command !== value.command ||
           prev.cwd !== value.cwd
         ) {
--- a/packages/app/src/agent/proc.test.ts
+++ b/packages/app/src/agent/proc.test.ts
@@ -61,16 +61,15 @@ describe("parseNetstatIb", () => {
 });
 
 describe("parsePsLine", () => {
-  // `ps -axo pid=,user=,pcpu=,pmem=,rss=,comm=` — rss is in KB on macOS,
+  // `ps -axo pid=,user=,pcpu=,rss=,comm=` — rss is in KB on macOS,
   // comm is last/greedy (may contain spaces).
   it("parses rss (KB) into absolute resident bytes", () => {
-    const parsed = parsePsLine("  501 alice 12.5  3.2 348160 /usr/bin/foo");
+    const parsed = parsePsLine("  501 alice 12.5 348160 /usr/bin/foo");
     expect(parsed).not.toBeNull();
     const [pid, proc] = parsed!;
     expect(pid).toBe(501);
     expect(proc.user).toBe("alice");
     expect(proc.cpuPct).toBe(12.5);
-    expect(proc.memPct).toBe(3.2);
     // 348160 KB × 1024 = absolute bytes (≈ 356 MB).
     expect(proc.rssBytes).toBe(348160 * 1024);
     expect(proc.command).toBe("/usr/bin/foo");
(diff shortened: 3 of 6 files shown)


srid:

| --- | --- | --- | --- |
| 1 | hickey | `Process.memPct` is now a dead wire field | Fixed in this PR |
| 2 | lowy | Vestigial per-process memPct field has no consumer | Fixed in this PR |
| 3 | hickey | Dropping memPct leaves dead darwin ps column + regex group | Fixed in this PR |

### Hickey

Hickey structural review of the host-detail diff (no new files added, so the duplication audit was skipped and the review run unprimed).

The bug-fix half of the diff is clean and needs no structural action. The Header's `pct`/`gb` were plain `const`s computed once at mount (when `props.system` is still `DEFAULT_SYSTEM`), so the memory readout froze at '0.0/0.0 GB (0%)'. Wrapping them in `createMemo` and reading `pct()`/`gb()` at the JSX sites is the correct fix and now mirrors the existing HostCard precedent (App.tsx:236). The replaced comment is accurate and the old misleading one is gone. No complecting introduced — `parsePsLine` is a clean extraction of pre-existing macOS parse logic (pure, unit-testable) rather than a new concept, so the concept-multiplication survey is not triggered.

The one structural concern is fragmentation in the feature half. The PR moved every live consumer of per-process memory to the new `rssBytes` field — the column header, the sort comparator (App.tsx:471), and the rendered cell (App.tsx:669) — but left `Process.memPct` in place across the schema, both agent readers, the delta diff loop, and the test. After this change `Process.memPct` is read by zero UI code: one domain concept ('physical memory this process occupies') now exists as two fields, one of which is dead. On Linux the two are a strict linear transform and cannot diverge; on macOS the agent still pays for the extra `pmem=` ps column whose output nothing consumes. The cost is not a correctness bug (nothing reads it, so nothing can break) but ongoing noise — an unnecessary computation per tick, an extra field in every transmitted frame, a redundant `!==` comparison in the hot diff over ~500 PIDs — plus a genuinely misleading schema comment ('memPct stays for sorting/ratio context') that contradicts the very PR that wrote it, since sorting was migrated to rssBytes here and no ratio consumer exists. I verified that all surviving `memPct` references in App.tsx/metrics.ts are the host-level `memPct(SystemInfo)` helper, not the process field, and that `processPctColor` is now correctly applied only to CPU.

Disposition is fix-in-PR: the dead field and its rationalizing comment were introduced/cemented by this PR, so there is no scope-based excuse to leave it. The removal is mechanical — surface.ts, proc.ts (both readers + LinuxProcRaw + stub + the darwin ps query drops `pmem=`), main.ts:197, and proc.test.ts:73 — and a future ratio consumer can derive `rssBytes / system.memTotal` at its call site since memTotal is already available wherever the process table renders. Keeping darwin/linux in parity (the stated goal) actually argues for the deletion: drop `pmem=` on macOS and stop computing the percentage on Linux, leaving both platforms reporting exactly the absolute bytes the UI now uses.

### Lowy

Reviewed `git diff origin/HEAD...HEAD` through the lowy volatility-based-decomposition lens. No files are added, so I skipped the duplication audit and reviewed unprimed.

The decomposition is fundamentally sound, and two things the research summary flagged as risks turned out to be non-issues worth recording:

1. No duplicate schema. The research summary warned of a 'DUPLICATE ProcessSchema redefined in agent proc.ts:19-28'. That is incorrect — proc.ts imports the `Process` type from `../common/surface` (proc.ts:29-37) and does not redefine the schema. The wire contract has a single source of truth, and adding `rssBytes` in one place (surface.ts) correctly flows to all readers via the derived type. This is the right boundary: the schema owns the shape, the platform readers own the volatile acquisition.

2. Volatility is well-encapsulated. The most volatile thing here is the darwin `ps` output format — a fragile, space-delimited string whose column order matters (comm must stay trailing/greedy). The diff extracts `parsePsLine` as a pure, exported function (proc.ts) and adds focused tests (proc.test.ts) covering the KB→bytes conversion, embedded-space commands, and malformed lines. Pulling that parsing behind a testable seam is a real encapsulation win: the format change (`pmem=,comm=` → `pmem=,rss=,comm=`) is now contained behind one function with one regex and one test, rather than inlined in the reader loop. Linux/darwin parity is symmetric (×1024 from kB on both), and the decimal-unit `formatBytes` reuse matches the established `memGb` convention, so memory and network sizes read consistently.

The bug fix is also a structural improvement, not just a behavioral one: replacing the two plain consts with `createMemo` (App.tsx:486) aligns the Header with the HostCard precedent (App.tsx:236), so both per-host views now derive memory through the same reactive mechanism. The replaced comment now states the truth (component body runs once at mount) instead of the prior false claim that it re-runs on prop change.

The single finding is the per-process `memPct` field, which this diff turns into a vestigial concept. Its sole production consumer (the sort comparator and the rendered MEM% cell) was migrated to `rssBytes`, so the field is now written across five sites and read by none. Carrying a value's volatility across the entire wire boundary when no module depends on it is accidental coupling — and the diff's own new schema comment ('memPct stays for sorting/ratio context') describes a usage that no longer exists. Either remove the field so the surface carries only consumed concepts, or fix the comment to reflect reality. I mark it 'fix' because there is no defer; it is low-severity and a reviewer optimizing for minimal change could legitimately keep the field, but the misleading justification should not stand.

srid:
## Evidence

Host-detail view at `/?host=localhost`, captured against a fresh dev server (linux, 487 PIDs live). One frame proves both changes flow schema → agent → wire → renderer.

host detail — header mem + absolute MEM column

**Bug fix — header mem is live, not frozen.** The header stat strip reads `mem 10.4/29.2 GB (36%)` — a real, polled value. Before the fix the header computed `pct`/`gb` as plain consts in the component body, so it rendered `0.0/0.0 GB (0%)` forever (the body runs once at mount when `currentSystem()` is still `DEFAULT_SYSTEM`). Now reactive, it tracks `load`/`uptime`/`os` like the rest of the strip.

**Feature — per-process MEM is now absolute, auto-scaled.** The process table column header is now `MEM` (verified `MEM%` no longer appears anywhere in the table). Cells render absolute resident memory via the decimal `formatBytes` helper, auto-scaling MB/GB:

| sample MEM cells |
| --- |
| 701.7 MB · 591.2 MB · 575.2 MB · 300.5 MB · 270.2 MB · 179.6 MB · 177.2 MB · 8.4 MB |

`rssBytes` is sourced from `VmRSS` (Linux, kB→bytes) and `ps -axo …rss…` (darwin, KB→bytes), kept in parity across both readers. The header now sorts and displays on the absolute value.

srid:

| sync | ✓ | 11s | Fetched origin; branch mem 0/0 vs upstream; dirty tree (untracked html) noted, continued |
| research | ✓ | 32s | Read all cited call sites (Header, HostCard, pidComparator, ProcessRow, formatBytes, schema, both readers, README) |
| branch | ✓ | 6s | Kept feature branch mem (not master) |
| implement | ✓ | 8m 20s | Both changes: reactive Header mem; absolute MEM column via rssBytes through schema + both readers + sort/render |
| check | ✓ | 1m 40s | bun run typecheck (tsc --noEmit) clean |
| docs | ✓ | 20s | README.md in sync (rssBytes + MB/GB column description) |
| fmt | ✓ | 8s | just fmt exit 0 |
| commit | — | 21s | d2476a6 pushed; tracked diffs empty |
| review | ✓ | 11m 7s | hickey+lowy: 3 findings, 3 fixed |
| police | ✓ | **16m 1s** | code-police: 3 violations, fixed |
| test | ✓ | 2m 33s | bun test: 45 pass; new parsePsLine block covers darwin rss KB→bytes path |
| revalidate | ✓ | 31s | Check + Test gates re-run green on HEAD |
| create-pr | ✓ | 1m 28s | PR #12 open; 14 CI checks pass; hickey/lowy ledger posted |
| ci | ✓ | 1m 50s | justci all 14 nodes green (x86_64-linux + aarch64-darwin) |
| evidence | ✓ | 2m 21s | Host-detail screenshot (live header mem, absolute MEM column) posted |
| **Total** | | **47m 29s** | |

1. **Quality gates dominated (~64% of wall time).** `police` (16m 1s) + `review` (11m 7s) = 27m 8s of the 47m 29s total. Both passes found exactly 3 issues each that were fixed in-loop — i.e. the implement step shipped 6 fixable issues. Front-loading a `code-police` self-pass during `implement` (or invoking it with the diff scoped to only the changed files) would catch these earlier and shrink the most expensive serial leg.

2. **CI was never the bottleneck — but it cost one avoidable re-run.** The first `ci` invocation failed solely because the untracked `talk-mem-header-bug.html` dirtied the tree under strict-mode HEAD pinning (justci pins HEAD). `sync` already flagged the dirty tree. Future runs: stash or remove stray untracked files at `sync` time so the first `justci` run is green and you skip the second invocation.

3. **`test` (2m 33s) re-ran the full suite when only proc.test.ts changed.** The new coverage is entirely in `parsePsLine`. For iteration speed, `--from`-style scoped re-runs (`bun test packages/app/src/agent/proc.test.ts`) during the inner loop, reserving the full `bun test` for the final `revalidate`, would have trimmed the test leg without losing the gate.

4. **No flaky retries and no wasted research.** research (32s) and revalidate (31s) were both lean and single-pass — no pre-reading or caching needed there. The gains are concentrated in the review/police serial chain (#1), not in setup.

PR #14, opened 2026-05-29
Persist per-host view prefs and add a light/dark toggle

srid:
**Your in-browser reading preferences now survive a reload.** Until now only the selected host persisted (in the URL) — the chart window, process sort column, and filter all snapped back to defaults on every refresh or tab switch, and the theme was whatever the OS dictated with no way to override it. This wires those preferences through `localStorage` and adds a light/dark toggle.

### What persists now

| Preference | Store | Scope |
| --- | --- | --- |
| Chart window (1m/5m/15m/30m), process sort column, process filter | `localStorage` | **per host** |
| Light/dark theme | `localStorage` | global |
| Selected host (unchanged) | URL `?host=` | — |

The split is deliberate. The URL stays the home for *shareable* state — a link you hand someone resolves to the same host. Window/sort/filter are *private* reading conveniences; putting them in a URL would be noise in a shared link. So they get their own store, in a new `localStorageState.ts` that mirrors `urlState.ts`'s shape: **the signal is the source of truth, the store is its projection** — pure namespaced read/write here, the `createSignal` seed + write-back effect in the component. Per-host scope falls out of where the signals already live: `HostView` is keyed-remounted per host, so each tab restores its own settings.

### Theme

The `dark:` Tailwind variant is rebound from the OS `prefers-color-scheme` media query to an explicit `data-theme` attribute, so a click can override the system theme. To avoid a flash of the wrong colours, the attribute is applied *before first paint* by a tiny inline bootstrap in `index.html` — it can't import the module (it runs before the module graph loads), so the storage key is the one duplication that can't be collapsed, guarded instead by a canary test.

> The history ring itself is still server-owned and in-memory (unchanged) — this PR only persists the *client's view* of it.

### Structural review

hickey + lowy ran on the diff (full analysis in a PR comment). Three findings landed as follow-up commits: the theme write-back was reshaped to match the signal-is-truth pattern, the three repeated per-host pref pairs collapsed into a `createPersistedSignal` helper, and the unavoidable key duplication got a canary test. Two were No-ops.

### Try it locally

```sh
nix run github:srid/drishti/localpersist-prefs
```

Then open <http://localhost:7720>, change the window/sort/filter on a host, toggle the theme top-right, and reload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Every view has its own URL, so you can bookmark or share a link straight to a ho
 
 Use the `+ add host` button in the tab strip to add hosts at runtime; the `×` on each tab removes one. Added/removed hosts persist to `$XDG_STATE_HOME/drishti/hosts.json` (override with `DRISHTI_HOSTS_FILE`), so `nix run github:srid/drishti` with no args restores the last session.
 
+Your in-browser reading preferences stick across reloads via `localStorage`: the **chart window** (1m / 5m / 15m / 30m), the **process sort column**, and the **process filter** are remembered **per host**, so each tab reopens the way you left it. A **light/dark toggle** (top-right of the tab strip) overrides the OS theme and is remembered globally; until you touch it, the theme follows your system preference.
+
 Requirements:
 
 - The remote host must be `ssh`-reachable with **passwordless** auth and a working **`nix-daemon`** that **trusts your user** (`trusted-users` in `nix.conf`) — drishti provisions the agent by shipping its `.drv` to the remote with `nix copy --derivation` and realising it there.
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -25,6 +25,7 @@ import {
   on,
   onCleanup,
   Show,
+  type Signal,
 } from "solid-js";
 import { createStore, reconcile } from "solid-js/store";
 import {
@@ -65,11 +66,37 @@ import {
 } from "../common/history";
 import { TabStrip } from "./TabStrip";
 import { prefKey, readPref, writePref } from "./localStorageState";
-import { applyTheme, initialTheme, otherTheme, type Theme } from "./theme";
+import {
+  applyTheme,
+  initialTheme,
+  otherTheme,
+  THEME_KEY,
+  type Theme,
+} from "./theme";
 import { adminClient, disposeHostSurface, surfaceForHost } from "./wire";
 
 const SORT_KEYS = ["cpu", "mem", "pid", "user"] as const;
 type SortKey = (typeof SORT_KEYS)[number];
+
+// Bind a per-host preference to a signal: seed from localStorage and mirror
+// every change back, keyed by host (`defer` skips the redundant write of the
+// just-read seed). Computing the storage key once is the point — it removes
+// the hazard of the seed-read and the write-back drifting to different keys.
+// This is the signal-is-truth / store-is-projection shape of urlState's URL
+// mirror, against a private store (localStorage) instead of the shareable
+// one (the URL). It lives here, not in the framework-free localStorageState.ts
+// (HostView is its only caller), so that module keeps its pure-I/O boundary.
+function createPersistedSignal<T extends string>(
+  pref: string,
+  host: string,
+  fallback: T,
+  accept?: (raw: string) => boolean,
+): Signal<T> {
+  const key = prefKey(pref, host);
+  const [value, setValue] = createSignal<T>(readPref(key, fallback, accept));
+  createEffect(on(value, (v) => writePref(key, v), { defer: true }));
+  return [value, setValue];
+}
 
 export default function App() {
   const admin = adminClient();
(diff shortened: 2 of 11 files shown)


srid:

Ran both lenses on the diff, then a cross-validation pass (each lens auditing the other's recommendations). The cross-pass mattered here — it flipped two first-pass recommendations.

| 1 | Hickey + Lowy | Theme write-back inverted the signal-is-truth pattern the rest of the change follows (DOM written before the signal) | Fixed in this PR |
| 2 | Lowy | Three per-host pref `signal + effect` pairs repeated verbatim; seed-key/write-key held together only by typing `prefKey` twice | Fixed in this PR |
| 3 | Hickey (cross-validation) | `THEME_KEY` literal duplicated in `index.html`'s pre-paint script with no enforcement | Fixed in this PR |
| 4 | Hickey (first pass) | Dedupe the `"drishti"` namespace by exporting `NAMESPACE` from the storage module into `theme.ts` | ⚠️ **No-op** |
| 5 | Hickey + Lowy (duplication audit) | Do the new `localStorageState.ts` / `theme.ts` reinvent the `urlState.ts` signal→store pattern? | ⚠️ **No-op** |
| 6 | Lowy + Hickey | `applyTheme` braided DOM mutation with the storage write | ⚠️ **No-op** |

### Hickey rationale

The duplication audit (#5) came back **clean**: `localStorageState.ts` faithfully follows the `urlState.ts` + `createEffect` pattern (pure read/write, signal seeded at the call site, mirrored via a deferred effect) — extension, not concept multiplication. `theme.ts` is a genuinely different primitive (it owns a DOM side-effect `urlState` never touches), so the split is justified.

The real bite was the theme path (#1): `toggleTheme` wrote the DOM via `applyTheme` *then* set the signal, making the DOM the functional source of truth and the signal a trailing tracker — and leaving the codebase with two competing templates for "persist a signal change." Fixed by making the signal the source and projecting it through a deferred effect.

On #4 (cross-validated to No-op): the cross-pass agreed with Lowy that exporting `NAMESPACE` couples two modules across a weak-volatility boundary for an *unachievable* DRY win — the `index.html` literal can't be eliminated regardless. Replaced with the canary test (#3), which guards the one duplication that genuinely can't be collapsed.

### Lowy rationale

Prior-encapsulation check (#5): no duplicated receptacle. `localStorageState.ts` shares the *pattern* of `urlState.ts` but encapsulates a different volatility axis — **shareable navigation state** (URL) vs **private per-browser reading preferences** (localStorage). Noise in a shared link is exactly why the URL is the wrong store for these; the module's own doc names this.

Finding #2: the *number* of per-host preferences is genuinely volatile (three today, a fourth plausible), so the repeated reactive wiring earned a named abstraction — `createPersistedSignal`. Cross-validation refined its **placement**: not inside the framework-free `localStorageState.ts` (that would braid solid-js reactivity into the pure storage layer), but as an App-local helper collocated with its sole caller, `HostView`.

#6 was judged a No-op only because it is **subsumed by the #1 fix** — reshaping the theme path leaves `applyTheme` as a DOM-only function (the storage write moved into the projection effect), which is exactly the decompose-the-braid outcome.

_Reviewers ran on Sonnet (skill frontmatter); cross-validation is a second invocation of each lens over the other's findings._

srid:
## Evidence

### Light/dark theme toggle

default
*Loads in **light** theme (OS preference); the ☀/☾ toggle sits top-right of the tab strip, labelled "Switch to dark theme".*

toggled
*After one click — the whole UI flips to **dark** theme; the button now reads "Switch to light theme", confirming the `data-theme` wiring works end-to-end.*

### Per-host preferences persist across reload

set
*15m window selected (green chip), MEM% sort active (▾ indicator on the MEM column), and `ssh` filter applied — showing 6 of 439 processes.*

persisted
*After a full page reload — 15m chip still highlighted, MEM ▾ sort still active, and `ssh` filter text still present (showing 5 of 441 processes — count shifted slightly as PIDs changed between polls, but filter and sort are intact). Dark theme also persisted, confirming all per-host state survives a reload via `localStorage`.*

srid:

| research | ✓ | 4m 13s | Mapped persistence (urlState only), per-host signal ownership in keyed HostView, theme wiring points, test conventions |
| branch | ✓ | 18s | On feature branch localpersist-prefs off origin/master |
| implement | ✓ | 3m 34s | Phase 1 (per-host filter/sort/window) + Phase 2 (theme toggle) implemented; tests added |
| check | ✓ | 16s | just typecheck (tsc --noEmit) exited 0 |
| docs | ✓ | 18s | README updated for per-host prefs + light/dark toggle |
| fmt | ✓ | 11s | just fmt (nixpkgs-fmt) clean |
| commit | ✓ | 19s | Primary feature commit 6df294c pushed |
| hickey+lowy | ✓ | **9m 52s** | hickey+lowy + cross-validation; 3 findings Fixed, 2 No-op (1329190, 41ac325) |
| police | ✓ | 4m 0s | rules + fact-check clean; elegance fixed otherTheme reuse (d8fc249) |
| test | ✓ | 13s | bun test: 67 pass, 0 fail |
| create-pr | ✓ | 1m 2s | Draft PR #14; hickey/lowy ledger posted |
| ci | ✓ | 1m 51s | justci: all 14 nodes succeeded (linux + darwin), covered HEAD d8fc249 |
| evidence | ✓ | 2m 41s | 4 screenshots posted: theme toggle + per-host prefs persisted across reload |
| **Total** | | **29m 10s** | |

**Slowest step**: `hickey+lowy` (9m 52s)

### Optimization suggestions

- **`hickey+lowy` dominated (9m 52s, 34% of total).** The cross-validation pass (a second round of both lenses auditing each other) roughly doubled review wall-clock — but it flipped two first-pass recommendations (rejected a `NAMESPACE` export, relocated `createPersistedSignal` out of the pure module), so it earned its cost here. On diffs with ≤1 finding, skipping cross-validation would reclaim ~4-5m.
- **`research` (4m 13s) re-derived what the `/talk --html` artifact already mapped.** This PR grew out of a talk-mode session whose artifact already located every signal and wiring point; feeding that artifact to `/do` as upfront context would trim research substantially.
- **CI's first run failed instantly** on the untracked `talk-localpersist.html` dirtying the tree (justci strict mode). Stashing or gitignoring stray non-PR artifacts before `/do` avoids the false-fail and the re-run.
- For any re-run from here, `--from ci-only` jumps straight to CI.

Workflow completed.

PR #18, opened 2026-05-29
Run the monitor as a service via home-manager (Linux + macOS)

srid:
**drishti now ships a home-manager module** that runs the monitor as a long-lived service — a systemd user service on Linux, a launchd LaunchAgent on macOS — selected automatically from the host platform. Consumers import one module, set `enable` + `package`, and never touch the supervisor distinction.

```nix
{
  imports = [ drishti.homeManagerModules.default ];
  services.drishti = {
    enable = true;
    package = drishti.packages.${system}.default;
    port = 7720;                  # default
    hosts = [ "user@host-a" ];    # optional; empty = manage hosts at runtime
  };
}
```

### One interface, two supervisors

The single axis of volatility — *which init system manages the process* — is encapsulated behind the stable `services.drishti` contract. The shared `args` / env facts are assembled once, then each platform branch converts them into its own transport shape:

```
services.drishti (enable, package, port, hosts, hostsFile)
        │
   ┌────┴─────────────────┐
 isLinux               isDarwin
 systemd.user.services  launchd.agents
 (Restart=on-failure)   (KeepAlive: Crashed + !SuccessfulExit)
```

The launchd `KeepAlive` is set to restart on both non-zero exit *and* crash signals, matching systemd's `Restart = "on-failure"` (plain `SuccessfulExit` would miss SIGSEGV and friends). On macOS the agent also writes `~/Library/Logs/drishti.{out,err}.log`, since launchd drops stdout/stderr by default.

### Subtracted from the kolu template, not copied

The module follows juspay/kolu's three-file pattern (drishti already builds on `@kolu/surface`), but the option surface is pared down to drishti's *actual* CLI (`packages/app/src/server/main.ts`): positional `[host...]` plus a single `--port`. kolu's `tls.*`, `verbose`, `diagnostics.*`, and `--host` are dropped — the server binds `0.0.0.0` hardcoded, and adding receptacles for non-existent flags would be speculative. The only env knob is `hostsFile` → `DRISHTI_HOSTS_FILE`; the packaged wrapper already self-contains `DRISHTI_DIST_DIR` / `DRISHTI_AGENT_DRVS_JSON` and the `openssh` + `nix` PATH wiring, so the module duplicates none of it.

### Tested in CI on both platforms

A new `home-manager` recipe in `ci/mod.just` builds `nix/home/example/` against this repo (`--override-input flake/drishti .`):

- **Linux** — a `nixosTest` boots a VM, waits for alice's user session, and curls the systemd service's port (UID resolved at runtime, port read back from the evaluated config).
- **Darwin** — a standalone home-manager activation build plus `assert`s on the generated launchd plist (no live launchd session needed on CI builders).

> The example's committed `flake.lock` pins `drishti` to the current master tip, which only exports `homeManagerModules` once this PR merges. The example header documents the `--override-input drishti .` workaround for standalone evaluation until then; `nix flake update` self-heals it post-merge.

### Try it locally

```sh
nix flake check github:srid/drishti/hm?dir=nix/home/example --override-input drishti github:srid/drishti/hm
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,28 @@ just regenerate-bun-nix           # after any bun.lock change
 
 The first connect to a fresh remote ships the agent closure over `ssh` (`nix copy --derivation` then `nix-store --realise`); subsequent connects reuse it. The progress is streamed to the browser via the `connection` cell.
 
+## Deployment (home-manager)
+
+A home-manager module runs the monitor as a systemd user service on Linux and as a launchd LaunchAgent on macOS:
+
+```nix
+{
+  imports = [ drishti.homeManagerModules.default ];
+  services.drishti = {
+    enable = true;
+    package = drishti.packages.${system}.default;
+    port = 7720;                  # default
+    hosts = [ "user@host-a" ];    # optional; empty = manage hosts at runtime
+  };
+}
+```
+
+The monitor binds `0.0.0.0` on `port`. `hosts` are passed as positional arguments; leave it empty to let drishti seed from its persisted hosts file (`$XDG_STATE_HOME/drishti/hosts.json`, overridable via `services.drishti.hostsFile`) and manage the set from the admin surface. The packaged wrapper already bakes `DRISHTI_DIST_DIR` / `DRISHTI_AGENT_DRVS_JSON` and puts `openssh` + `nix` on `PATH`, so the service self-contains its runtime needs.
+
+See [`nix/home/example/`](nix/home/example/) for a full configuration — a NixOS VM test exercises the systemd path on Linux, and a standalone home-manager activation build exercises the launchd path on Darwin.
+
+On macOS, the LaunchAgent writes stdout to `~/Library/Logs/drishti.out.log` and stderr to `~/Library/Logs/drishti.err.log`, so crashes and startup failures leave logs alongside other user logs.
+
 ## Project layout
 
 ```
@@ -110,9 +132,12 @@ drishti/
 │  ├─ nixpkgs.nix
 │  ├─ overlay.nix             # kolu-surface, kolu-surface-nix-host
 │  ├─ env.nix                 # DRISHTI_KOLU_SURFACE{,_NIX_HOST}
-│  └─ packages/
-│     ├─ kolu-package.nix     # mkKoluPackage factory
-│     └─ drishti/default.nix  # bun2nix build derivation
+│  ├─ packages/
+│  │  ├─ kolu-package.nix     # mkKoluPackage factory
+│  │  └─ drishti/default.nix  # bun2nix build derivation
+│  └─ home/
+│     ├─ module.nix           # home-manager module (systemd / launchd)
+│     └─ example/             # example config — CI-built (VM + launchd checks)
 ├─ scripts/
 │  └─ hydrate-kolu-packages.sh
 └─ packages/app/
--- a/ci/mod.just
+++ b/ci/mod.just
@@ -15,7 +15,7 @@ nix_shell := if env('IN_NIX_SHELL', '') != '' { '' } else { 'nix develop --accep
 [macos]
 [parallel]
 [metadata("ci")]
-default: nix typecheck fmt-check bun-nix-fresh
+default: nix typecheck fmt-check bun-nix-fresh home-manager
 
 # Single per-platform `bun install`. Funnel every bun consumer through this
 # node so concurrent `bun install` runs against the same workspace don't race.
@@ -38,6 +38,13 @@ _prefetch-crates:
 # Build all flake outputs via devour-flake.
 nix: _prefetch-crates
     nix build github:srid/devour-flake -L --no-link --print-out-paths --override-input flake .
+
+# Build the home-manager example flake (nix/home/example) against THIS repo:
+# the Linux nixosTest boots a VM and curls the systemd user service, and the
+# Darwin check eval-builds the launchd plist. `flake/drishti` is overridden to
+# the repo root so the example consumes the local module, not the published one.
+home-manager: nix
+    nix build github:srid/devour-flake -L --no-link --print-out-paths --override-input flake ./nix/home/example --override-input flake/drishti .
 
 # TypeScript type checking
 typecheck: install
--- a/flake.nix
+++ b/flake.nix
@@ -68,6 +68,11 @@
       # system attr structure.
       agentDrvsJson = builtins.toJSON agentDrvBySystem;
 
+      # home-manager module — runs the monitor as a systemd user service on
+      # Linux and a launchd LaunchAgent on macOS. System-independent; the
+      # consumer supplies `services.drishti.package` per their system.
+      homeManagerModules.default = import ./nix/home/module.nix;
+
       # `nix fmt` — format *.nix files only.
       formatter = eachSystem ({ pkgs, ... }: pkgs.nixpkgs-fmt);
 
--- a/nix/home/example/flake.lock
+++ b/nix/home/example/flake.lock
@@ -1,0 +1,164 @@
+{
+  "nodes": {
+    "bun2nix": {
+      "inputs": {
+        "flake-parts": "flake-parts",
+        "nixpkgs": "nixpkgs",
+        "systems": "systems",
+        "treefmt-nix": "treefmt-nix"
+      },
+      "locked": {
+        "lastModified": 1779931262,
+        "narHash": "sha256-8ypPqVPCxiLD6GAvGjA+cJ3REJzgMJDGkGJLZYuCfGo=",
+        "owner": "juspay",
+        "repo": "bun2nix",
+        "rev": "ce67cc1885e5e28fcc3ff865479be1a19ff396cb",
+        "type": "github"
+      },
+      "original": {
+        "owner": "juspay",
+        "ref": "rawflake",
+        "repo": "bun2nix",
+        "type": "github"
+      }
+    },
+    "drishti": {
+      "inputs": {
+        "bun2nix": "bun2nix"
+      },
+      "locked": {
+        "lastModified": 1780084001,
+        "narHash": "sha256-X4P4AYHm7hDdd8jw3/uZyCB5Y9tCzYlAUtn6b6AC1wA=",
+        "owner": "srid",
+        "repo": "drishti",
+        "rev": "6c06b5ff45a3a3bd4098fbc87c94ebcbef0c2f87",
+        "type": "github"
+      },
+      "original": {
+        "owner": "srid",
+        "repo": "drishti",
+        "type": "github"
+      }
+    },
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": [
+          "drishti",
+          "bun2nix",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1777988971,
+        "narHash": "sha256-qIoWPDs+0/8JecyYgE3gpKQxW/4bLW/gp45vow9ioCQ=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "0678d8986be1661af6bb555f3489f2fdfc31f6ff",
+        "type": "github"
+      },
+      "original": {
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "type": "github"
+      }
+    },
+    "home-manager": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1779969295,
+        "narHash": "sha256-HwIJ3tOcwSMiV75L7KqJXciXR9UfT+d7rwOZMX7cTnA=",
+        "owner": "nix-community",
+        "repo": "home-manager",
+        "rev": "61e2c9659324181e0f0ed911958c536333b1d4f6",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "home-manager",
+        "type": "github"
+      }
+    },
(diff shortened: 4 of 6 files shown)


srid:

| --- | --- | --- | --- |
| 1 | hickey | Duplication audit: new home-manager surface, not a reinvention | ⚠️ **No-op** |
| 2 | hickey | Committed example flake.lock pins drishti to a pre-PR commit lacking homeManagerModules | Fixed in this PR |
| 3 | hickey | `--port` always passed even at default value | ⚠️ **No-op** |
| 4 | hickey | Platform supervisor selection is cleanly disjoint | ⚠️ **No-op** |
| 5 | hickey | hostsFile str type leaks absolute-path invariant to prose | Fixed in this PR |
| 6 | hickey | Redundant nixosConfigurations.example duplicates the VM-test build path | Fixed in this PR |

### Hickey Review — `hm` branch (home-manager module)

#### Scope correction

The actual PR diff (`git diff origin/HEAD...HEAD`) is purely the home-manager addition: `README.md`, `ci/mod.just`, `flake.nix`, and three new files under `nix/home/`. (The skill's forked pass described charts/theme/NIC features — those belong to prior commits already on `master`, not this branch — so I reviewed the real diff directly.)

#### Duplication audit (new files present)

The diff adds `nix/home/module.nix` and `nix/home/example/flake.{nix,lock}`. There is no pre-existing home-manager module in drishti, so this is new surface rather than a reinvention of an in-repo pattern. It mirrors the upstream **kolu** three-file pattern (drishti already depends on kolu as its surface library) and is correctly adapted to drishti's real CLI: kolu's `tls.*`, `diagnostics.*`, `host`, and `verbose` are dropped; only `port` (default 7720), positional `hosts`, and a `hostsFile` → `DRISHTI_HOSTS_FILE` override remain — matching `packages/app/src/server/main.ts` (cleye: `[host...]` + `--port`, binds `0.0.0.0` hardcoded). The packaged wrapper already bakes `DRISHTI_DIST_DIR`/`DRISHTI_AGENT_DRVS_JSON` and prefixes `openssh`+`nix` onto PATH, so the module correctly sets no extra env or PATH wiring. No headline duplication finding.

#### Structural quality

The module is a clean decomposition. The `args` list and `envAttrs` are pure values; the two supervisor blocks (`systemd.user.services` under `isLinux`, `launchd.agents` under `isDarwin`) are disjoint, each converting the shared facts into its own transport shape at the call site. One fact expressed two ways, no interleaving. I verified: `nixpkgs-fmt --check` passes on both new nix files; with `--override-input drishti <repo>` (as CI does) the example evaluates and `ExecStart` resolves to `drishti --port 7720`. The README "Deployment (home-manager)" section and "Project layout" tree are accurate (the `kolu-*` tree entries are pre-existing and untouched; only the `home/` subtree was added).

#### The one fix

**The committed example `flake.lock` pins `drishti` to `6c06b5f` — the master tip *before* this PR — which does not export `homeManagerModules`.** Evaluating the example standalone (without the CI `--override-input flake/drishti .`) fails with `error: attribute 'homeManagerModules' missing` (verified). CI is green only because `ci/mod.just`'s `home-manager` recipe supplies that override. This complects the example's evaluability with an out-of-band flag living in another file: the header comment calls it a "full configuration" reference, but as committed it is not buildable on its own and only self-heals once this commit reaches `srid/drishti` and a consumer runs `nix flake update`. Lowest-friction remedy: note in the example header that standalone evaluation needs `--override-input drishti <repo>` until the module lands upstream, and regenerate/refresh the lock against the merge commit.

Everything else (always-passed `--port`, the platform `mkIf` split) is correct as written and marked no-op.

### Lowy review — home-manager module for drishti (branch `hm`)

**Duplication audit (required — diff adds new files).** The diff adds `nix/home/module.nix`, `nix/home/example/flake.{nix,lock}`. I audited the repo for a canonical in-repo decomposition of the same KIND (a home-manager service module / process-supervisor abstraction). `find nix -type f` and a grep for `homeManagerModules`, `launchd.agents`, `systemd.user.services`, and `mkEnableOption` across all `.nix` files turned up nothing outside `nix/home/` — the only other `homeManagerModules` reference is the new `flake.nix` export wiring the module in. This is the first home-manager module in drishti: genuinely new work, not a reinvention of an existing in-repo pattern. The audit clears with no headline finding.

**Boundaries examined.**
1. `services.drishti` option block in `nix/home/module.nix` — the stable consumer-facing contract.
2. The `isLinux`/`isDarwin` `config` split (systemd.user.services vs launchd.agents) — the activity-volatility axis the module exists to encapsulate.
3. The `args` and `envAttrs` let-bindings — shared CLI-arg / env assembly consumed by both branches.
4. `homeManagerModules.default` export in `flake.nix` — the public boundary consumers import.
5. `nix/home/example/flake.nix` — CI test harness, not a production decomposition decision.

**Volatility map.** The single real axis of volatility — *which supervisor manages the process* (systemd on Linux, launchd on macOS) — is fully encapsulated behind the stable `services.drishti` interface. Consumers express intent (`enable`, `package`, `port`, `hosts`, `hostsFile`) and never touch the supervisor distinction; the platform branch is selected internally via `pkgs.stdenv.hostPlatform.isLinux/isDarwin`. This is a textbook activity-volatility encapsulation: one receptacle, one axis, consumer shielded.

The `args` binding co-locates exactly one axis (drishti's CLI surface) and is verified against `packages/app/src/server/main.ts` — positional `[host...]` plus a single `--port` flag, no `--host`/`--tls`/`--verbose`. The port server-binds `0.0.0.0` hardcoded, so the module correctly exposes `port` but not `host`. The `envAttrs` binding co-locates one axis (env config), surfacing only `DRISHTI_HOSTS_FILE` — a pre-existing, documented, test-backed env var (`hostsStore.ts`), not a speculative dimension.

**Correct subtraction, not verbatim copy.** The port correctly drops kolu's `tls.*`, `verbose`, and `diagnostics.*` axes (and the paired-cert assertion) because drishti's CLI has no such flags — adding option receptacles for non-existent axes would be speculative complexity. The module also correctly omits any `DRISHTI_DIST_DIR` / `DRISHTI_AGENT_DRVS_JSON` / PATH wiring, because drishti's wrapper (`default.nix`) self-contains those via `--set`/`--set-default`/`--prefix`. The package is the encapsulation point for its own runtime closure; the module does not duplicate that boundary. This is the right layering direction.

**`toString args` / always-passing `--port`.** The `ExecStart = toString args` shape and the unconditional `--port` (vs kolu's conditional construction) are inherited from the settled upstream kolu pattern; host identifiers (`user@host`) carry no whitespace, so word-splitting is not a boundary concern. Passing the default port explicitly is a benign single-axis choice, not a decomposition smell.

**Naming, symmetry, layering, expendability.** Naming is consistent (`drishti` throughout). The two supervisor branches are symmetric (same Description, Restart-equivalent KeepAlive semantics, same env-injection idiom). The example flake is an almost-expendable test harness — deleting it would lose CI coverage but not production behavior, which is the correct shape for a test boundary. README and `ci/mod.just` updates are documentation/CI plumbing, not new abstractions.

**Findings:** None. The module passes every volatility-decomposition check, and the duplication audit confirms no in-repo pattern is being reinvented.

srid:

| sync | ✓ | 16s | Ran git fetch origin and git remote set-head origin --auto (origin/HEAD unchanged, points to master). Branch hm tracks origin/master with 0 behind and 0 ahead, so no git pull --ff-only was needed. Working tree clean (git status --porcelain empty), so no dirty-tree hint emitted. Forge detected as github from https://github.com/srid/drishti.git. No branches/commits/PRs created. |
| research | ✓ | 50s | Read kolu's reference home-manager module (/tmp/kolu/nix/home/module.nix), its example flake with NixOS VM + Darwin launchd checks (/tmp/kolu/nix/home/example/flake.{nix,lock}), flake wiring (homeManagerModules.default), README section, and CI recipe. Cross-referenced drishti's actual CLI/env surface (packages/app/src/server/main.ts, hostsStore.ts), its flake.nix outputs, default.nix package attrs, ci/mod.just, and README. drishti has NO existing home module (confirmed via find nix -name '*.nix'). drishti's invocation surface differs from kolu's: positional [host...] args + --port flag (default 7720; also reads PORT env), DRISHTI_HOSTS_FILE env override, listens on 0.0.0.0; no --host/--tls/--verbose/diagnostics flags. Recorded start (1780084116) and end (1780084166) epochs. |
| branch | ✓ | 3s | HEAD is on feature branch 'hm' (tracking origin/master), not the default branch 'master'; no new branch needed. |
| implement | ✓ | 3m 10s | Added home-manager module (nix/home/module.nix: systemd user service on Linux, launchd agent on macOS), CI-built example flake with Linux VM test + Darwin launchd checks, flake.nix homeManagerModules.default output, ci/mod.just home-manager recipe + pipeline entry, and README section/layout updates. Verified: nixpkgs-fmt clean, Darwin launchd-config asserts eval to a .drv, Linux systemd ExecStart resolves to the drishti wrapper with --port 7720, and .#homeManagerModules = [ "default" ]. |
| check | ✓ | 42s | Ran \`bun run typecheck\` (tsc --noEmit) via nix dev shell; initially failed with TS2307 cannot find module 'cleye' due to stale node_modules, fixed via \`bun install\`, now green with no errors |
| docs | ✓ | 55s | README.md home-manager section matches code: module options (enable/package/port=7720/hosts/hostsFile), 0.0.0.0 bind + positional [host...] CLI (main.ts), DRISHTI_HOSTS_FILE + \$XDG_STATE_HOME/drishti/hosts.json (hostsStore.ts), baked DRISHTI_DIST_DIR/DRISHTI_AGENT_DRVS_JSON + openssh/nix on PATH + mainProgram=drishti (default.nix wrapper), macOS log paths and CI VM/launchd checks (module.nix, example/flake.nix). Project-layout tree updated for nix/home/. |
| fmt | ✓ | 6s | Ran format command \`just fmt\` (nixpkgs-fmt) from .agency/do.md; exited 0 with no errors |
| commit | ✓ | 17s | New commit 90232d7 "feat: add home-manager module for Linux and macOS" created on branch hm and pushed; local HEAD matches origin/hm (90232d7) |
| review | ✓ | **17m 30s** | hickey+lowy: 6 findings, 3 fixed |
| police | ✓ | **21m 27s** | code-police: 4 violation(s), fixed |
| test | ✓ | 20s | Inspected the branch diff (git diff origin/HEAD...HEAD): changed files are README.md, ci/mod.just, flake.nix, nix/home/module.nix, nix/home/example/flake.nix, nix/home/example/flake.lock — all Nix/docs, zero TypeScript/JS source (grep for .ts/.tsx/.js/.jsx/.mjs/.cjs returned no matches). The configured test command \`bun test\` exercises only TypeScript code under packages/app/src/{client,agent,server,common}, none of which this branch touches, so it would not cover the new behavior. The new home-manager module's behavior is verified instead by the dedicated \`home-manager\` recipe added to ci/mod.just (NixOS VM nixosTest that boots the systemd user service and curls http://127.0.0.1:7720/ on Linux; launchd plist eval-build with assertions on KeepAlive/StandardOutPath/StandardErrorPath on Darwin). The diff is exempt from \`bun test\`; relevant verification lives in the CI lane, not the unit-test command. |
| revalidate | ✓ | 39s | Re-ran configured gates on HEAD ee9fda4 via \`nix develop\`: \`bun run typecheck\` (tsc --noEmit) passed clean; \`bun test\` reported 77 pass / 0 fail across 10 files (the malformed-JSON log line is an intentional negative-path test, not a failure). Working tree clean, no fixes required. |
| create-pr | ✓ | 1m 14s | Draft PR #18 created (gh pr view succeeds, isDraft=true); title/body match the home-manager-only scope from the diff; Hickey/Lowy analysis posted as a PR comment with the findings ledger (No-op rows as warning-emoji bold) plus both lenses' rationale prose. |
| ci | ✓ | 3m 25s | justci ran \`nix run github:juspay/justci -- run --host x86_64-linux=localhost\` against HEAD ee9fda4 (clean tree); all 16 nodes succeeded on x86_64-linux and aarch64-darwin, including the home-manager recipe. |
| evidence | — | 16s | PR evidence section present but conditional on visible UI impact; PR #18's master..hm commits touch only Nix module/example/README/CI (no UI surface), so per the section's judgment guidance no Evidence comment is warranted. |
| **Total** | | **51m 10s** | |

1. **Review + police dominated the run (review 17m30s + police 21m27s = 38m57s, 76% of the 51m10s total).** Everything else combined finished in ~12m. For a Nix-only diff (module.nix + example flake + CI recipe + README), the structural-simplicity lenses (hickey/lowy) and code-police have limited TypeScript surface to act on. Consider scoping these passes — e.g. a lighter review tier for Nix/docs-only diffs, or running hickey/lowy and code-police concurrently rather than sequentially, which alone could reclaim ~17m.

2. **`check` lost time to stale `node_modules` (TS2307 cannot find module 'cleye'), needing a recovery `bun install`.** Add `bun install --frozen-lockfile` to the dev-shell entry or a pre-`check` hook so the typecheck never trips on stale deps. This was the only step that hit an error-and-retry.

3. **`test` (20s) was a no-op for this diff — `bun test` exercises only TypeScript under `packages/app/src/**`, which this branch never touches.** The real verification lived in the `home-manager` CI recipe (NixOS VM + Darwin launchd). For Nix/docs-only branches, the pipeline could skip the `bun test` lane (diff-aware gating) and lean directly on `ci`, since `revalidate` already re-runs `bun test` anyway.

4. **`ci` (3m25s) re-ran the full 16-node pipeline including both platforms after `revalidate` had already passed the unit gates.** Since the change is isolated to the new `home-manager` recipe, a future re-run could target just that recipe (`home-manager@x86_64-linux` / `@aarch64-darwin`) via `--from ci-only` to avoid rebuilding unrelated nodes.

PR #19, opened 2026-05-29
Make drishti strictly read-only by removing process kill

srid:
**drishti is a monitor, and now it can only ever observe — never act.** The single mutating capability was `process.kill` (send `TERM`/`KILL`/`HUP`/`INT` to a PID); this PR excises it end-to-end so there is no path from the UI or the wire to mutate a monitored host.

The kill capability threaded cleanly through one RPC procedure across every tier, so removal is a clean cut rather than a refactor:

```
surface schema  ─▶  agent handler  ─▶  parent router  ─▶  client UI
(process.kill)      process.kill()     forwards to agent   "kill" button
   removed             removed             removed            removed
```

The per-host surface now declares **cells, collections, and streams but no procedures at all** — `procedures` is optional in the surface spec, so dropping it leaves a well-formed read-only contract. The only procedures left anywhere live on the separate *admin* surface (host-set add/remove), which never touches a monitored host.

A guard test (`common/surface.test.ts`) asserts the surface exposes no procedures, so the read-only invariant can't be silently undone by a future change.

### Try it locally

```sh
nix run github:srid/drishti/nokill
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Open <http://localhost:7720>. The UI opens on the **fleet** tab — a single ove
 - **Runtime host management** — the `+ add host` button adds hosts, the `×` on each tab removes one. Added/removed hosts persist to `$XDG_STATE_HOME/drishti/hosts.json` (override with `DRISHTI_HOSTS_FILE`), so `nix run github:srid/drishti` with no args restores the last session.
 - **Per-host view memory** — the chart window, process sort column, and process filter stick across reloads via `localStorage`, remembered per host. A **light/dark toggle** (top-right of the tab strip) overrides the OS theme and is remembered globally; until you touch it, the theme follows your system preference.
 - **Idle NICs collapse** behind a `+N idle` toggle by default, so the few interfaces moving traffic aren't buried under the dozens of always-zero virtual ones (utunN, anpiN, …).
-- **Process kill** — sends `TERM` / `KILL` / `HUP` / `INT`; the only mutation.
+- **Strictly read-only** — drishti only ever *observes* a host. The per-host surface exposes no procedures, so there is no way to signal, kill, or otherwise mutate a monitored process through the UI or the wire.
 
 Requirements:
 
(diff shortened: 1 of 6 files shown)


srid:

| research | ✓ | 3m 46s | Mapped all kill touch-points across surface/agent/router/client/README |
| branch | ✓ | 0s | Feature branch `nokill`, level with origin/master |
| implement | ✓ | 1m 57s | Excised `process.kill` from surface/agent/router/client; updated README; added read-only guard test |
| check | ✓ | 56s | `just typecheck` (tsc --noEmit) passed |
| fmt | ✓ | 6s | `just fmt` ran without error |
| commit | ✓ | 12s | commit `5b3017e`, pushed to origin/nokill |
| docs | — | 0s | skipped via `--minimal` |
| hickey+lowy | — | 0s | skipped via `--minimal` |
| police | — | 0s | skipped via `--minimal` |
| test | ✓ | 1m 6s | `bun test`: 79 pass / 0 fail; new `surface.test.ts` covers the read-only invariant |
| create-pr | ✓ | 21s | Draft PR #19 created |
| **ci** | ✓ | **3m 56s** | justci: all 14 nodes succeeded (x86_64-linux@localhost + aarch64-darwin) |
| **Total** | | **12m 50s** | |

**Slowest step**: `ci` (3m 56s)

### Optimization suggestions

- **`ci` dominated** (3m 56s / 31% of total) — almost entirely Nix derivation builds (`ci::nix`, `bun-nix-fresh`). For re-runs after a trivial push, `--from ci-only` skips straight back here; the build cost is unavoidable since the closure changed.
- **`research` was the second-largest slice** (3m 46s). This change was a clean single-procedure excision — pre-identifying `surface.ts` / `agent/main.ts` / `router.ts` / `App.tsx` as the touch-points before invoking `/do` would have cut most of the exploration.
- **`test` ran the full 79-test suite** (1m 6s) when only `common/surface.test.ts` was relevant to the diff; scoping to the changed path would shave the run, though the full pass was cheap insurance here.

PR #20, opened 2026-05-29
Show a 30m CPU/memory history sparkline on each fleet card

srid:
The **fleet overview previously showed only instantaneous numbers** per host — CPU, memory, load, uptime — so a card told you "now" but never "trending up or down?". Each host card now carries a compact **30-minute CPU/memory sparkline**, drawn from the same parent-owned metric ring the full host view already uses. The overview becomes a true pane-of-glass: trend at a glance, no drill-down required.

The card pins the **widest** window (30m today, derived from `HISTORY_WINDOWS` so it tracks the retention bound) and deliberately omits the per-host duration picker — a card is a glanceable trend, not an interactive chart.

### How it reuses the existing chart

Rather than copy the host view's streaming + projection into the card, the two shared steps were lifted into reusable primitives so the call sites can't drift:

```
metricHistory stream ─▶ subscribeMetricHistory ─▶ projectHistory ─▶ Sparkline
                         (ring signal + error)     (polylines/window)   (SVG)
```

- **`subscribeMetricHistory`** — accumulates the streamed ring (snapshot + per-tick deltas) into a signal, off a caller-supplied `AbortSignal`.
- **`projectHistory`** — projects a ring to the chart's cpu/mem polyline strings + latest sample for a given window.
- **`Sparkline`** — the bare two-polyline SVG, pulled out of the existing `HistoryChart`; the caller sets its height (`h-24` in the host view, `h-10` on the card).

`HostView` was refactored onto these same primitives, so the full chart and the card sparkline share one code path.

### Refinements during structural review

| Lens | What it caught | Resolution |
| --- | --- | --- |
| Hickey | `subscribeMetricHistory` collapsed "no samples yet" and "stream died" into one empty ring — the lone per-host subscription swallowing failure to the console | Returns `{ history, streamError }`; the sparkline now shows a distinct `unavailable` vs `collecting…` overlay |
| Hickey → Lowy cross-val | A proposed "extract a shared legend component" would have braided the host view's *data-bound* legend (live %) with the card's *static* color key behind a mode flag | Left as-is (both lenses converged: the merge complects) |
| Police (altitude) | The card's `30m` span was a hardcoded string in two places | Derived `WIDEST_HISTORY_WINDOW` from `HISTORY_WINDOWS.at(-1)`, mirroring how `HISTORY_RETENTION_MS` is derived |

> _History remains in-memory only — the card sparkline inherits the same zero-config, parent-owned ring; restarting the parent starts fresh._

### Try it locally

```sh
[...]

PR #22, opened 2026-05-29
fix(fleet): sparkline background ignores dark mode

srid:
## What

The 30m CPU/memory sparkline (added in #20) renders with a **light** background in dark mode — on both the fleet card and the open host view. The trace lines are themed, but the box behind them stays `bg-gray-100`.

before — light sparkline boxes in a dark UI

## Why

The `Sparkline` background was written inside a template literal with the dark utility jammed directly against `${...}` interpolation:

```tsx
class={`... bg-gray-100 dark:bg-gray-800/50${props.class ? ` ${props.class}` : ""}`}
//                       ^^^^^^^^^^^^^^^^^^^^^ no separator before ${
```

Tailwind v4's static class extractor only sees **complete, statically-present** class names. A utility butted against `${` isn't recognized, so `.dark:bg-gray-800/50` was **never emitted into the compiled CSS** — only the base `bg-gray-100` survived, hence the light box.

The card's own `dark:bg-gray-900/40` lives in a plain `class="..."` string, so it compiled fine — which is why *only* the sparkline was affected, and why it looked like a one-off styling miss rather than a build-time drop.

I confirmed this against the real `@tailwindcss/cli` build:

| utility | context | emitted? |
|---|---|---|
| `dark:bg-gray-900/40` (card) | `class="..."` | ✅ |
| `dark:bg-gray-800/50` (sparkline, before) | `dark:bg-gray-800/50${...}` | ❌ |
| `dark:bg-gray-800/50` (sparkline, after) | `dark:bg-gray-800/50 ${...}` | ✅ |

## Fix

Keep the utility a statically-extractable token by putting a space before the interpolation:

```tsx
class={`... bg-gray-100 dark:bg-gray-800/50 ${props.class ?? ""}`}
```

One-line change. Both callers (`h-10` fleet card, `h-24` host view) pass a static `class`, so the `?? ""` form is equivalent. Verified the class is now present in the compiled CSS, and grepped the client for any other Tailwind utilities adjacent to `${` — none remain.

## Avoiding this in future

The trap: Tailwind can only generate CSS for class names it can find as literal substrings. Any time a class string is *built* (interpolation, concatenation, computed keys), a needed utility can silently vanish from the bundle — no error, just missing styles.

Cheap guardrails, roughly in order of bang-for-buck:

1. **Convention:** never butt a utility against `${`. Write `\`base… ${dynamic ?? ""}\`` (space + nullish), or skip the template entirely when the dynamic part is conditional.
2. **Prefer Solid's `classList`** for conditional classes and keep the base classes in one plain static string — the base set then never passes through interpolation at all.
3. **Tooling:** the official Prettier `prettier-plugin-tailwindcss` / VS Code Tailwind IntelliSense only act on statically-detectable classes, so a utility that "won't sort / won't autocomplete" is an early warning it won't compile either. `eslint-plugin-tailwindcss` can flag this in CI.
4. If a dynamic class genuinely can't be made static, **`@source inline(...)` / a safelist** is the escape hatch — but that's a last resort, not the default.

### DIFF
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -1136,7 +1136,7 @@ function Sparkline(props: {
 }) {
   return (
     <div
-      class={`relative w-full overflow-hidden rounded bg-gray-100 dark:bg-gray-800/50${props.class ? ` ${props.class}` : ""}`}
+      class={`relative w-full overflow-hidden rounded bg-gray-100 dark:bg-gray-800/50 ${props.class ?? ""}`}
     >
       <svg
         class="h-full w-full"


srid:
## Evidence

Both captured in **dark mode** on the fleet overview. The fix is visible in the `30M` sparkline box at the bottom of the host card — it was rendering with a light/white background (the dropped `dark:bg-gray-800/50`), and now renders dark to match the card.

**Before** (commit `f873b66`, buggy) — sparkline box is white, clashing with the dark card:

Before: sparkline has a white background in dark mode

**After** (this PR) — sparkline box is dark, matching the card:

After: sparkline has a dark background in dark mode

Issue #21, opened 2026-05-29
Per-process CPU/mem sparkline in the process info panel


srid:
## Context

Follow-up to the process info panel (phases 1 + 2). Clicking a process opens a detail panel showing its fields. Phase 3 adds a **per-process CPU/mem history sparkline** to that panel — a small trace of how the selected process's CPU% and RSS have moved over the recent window, mirroring the host-level `HistoryChart`.

## Why this is a separate (larger) phase

It's a backend lift, not a UI add. There is **no per-PID history today**:

- `metricHistory` is **system-wide** — `MetricSample` is `{t, cpu, mem}` with no PID (`packages/app/src/common/surface.ts:139-163`).
- The parent owns one in-memory ring per *host*, sampled every poll tick.

To draw a per-process trace we need per-PID retention. Sketch of the work:

- **Parent**: retain a bounded ring of `{t, cpuPct, rssBytes}` per live PID (evicted when the PID exits, to cap memory across 500+ short-lived processes). This is the bulk of the cost — sizing/eviction policy matters.
- **Surface**: a per-PID history stream (or extend `processesSnapshot`), keyed by PID, snapshot-then-delta like the existing rings.
- **UI**: render a sparkline in the panel, reusing `HistoryChart`'s polyline projection.

## Acceptance

- Selecting a process shows a CPU% and mem sparkline over the recent window.
- Parent memory stays bounded with high process churn (short-lived PIDs don't leak rings).
- Survives the panel's existing PID-exit handling (selection clears when the process dies).

Deferred from the info-panel design; phases 1 (panel over current fields) and 2 (enrich the surface with parent/state/nice/start/threads/argv) ship first.

PR #9, opened 2026-05-29
Give every host its own shareable URL

srid:
**Every view now has an address.** Selecting a host writes it into the browser URL as `?host=<ssh-target>` (the fleet overview is the bare `/`), so you can bookmark a host, paste a link in chat, or reload and land right back on the host you were watching — previously the app always reopened on the fleet tab regardless of where you'd navigated.

The existing `View` sum (`{kind:"fleet"} | {kind:"host";host}`) was already the perfect thing to serialize, so no router library was added. A small pure module does the encoding both ways; `App.tsx` just wires it to the view signal.

```
URL (?host=…)  ◀──replaceState──  view signal  ──init/restore──▶  URL
                                      │
                                 resolvedView ──▶ pane (fleet | host htop)
```

### How it works

- **`urlState.ts`** — two pure functions, `viewFromSearch` / `searchForView`, that round-trip a `View` through a `location.search` string. ssh targets are URL-encoded, so `user@host` survives intact. Unit-tested in isolation (no DOM), matching the `metrics` / `usageColors` pattern.
- **`App.tsx`** — the `view` signal initializes from the URL (a deep-linked host resolves once the admin collection arrives), and a `createEffect` mirrors changes back via `replaceState`. The URL binds to *intent*, not `resolvedView`, so a still-loading deep-link isn't rewritten to fleet before its host shows up.

### Edge cases handled

| Case | Behavior |
| --- | --- |
| Removing the host you're viewing | Intent resets to fleet so the URL clears, not just the pane |
| Malformed `?host=` / `?host=%20` | Trims to blank → falls back to fleet instead of sticking forever |
| Deep-link to a host still loading | URL preserved as intent; resolves when the admin collection arrives |

> _`replaceState` (not `pushState`) keeps the URL reflecting the current view without growing the back-button stack with programmatic corrections._

### Try it locally

```sh
nix run github:srid/drishti/url
```

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

PR #23, opened 2026-05-29
Click a process row to open a detail panel

srid:
**Clicking a row in the process table now opens an inline detail panel** above the table, showing the selected process in full: the untruncated command and working directory the table clips, exact CPU% and resident memory (plus its share of host RAM), and — new to the `Process` surface — parent PID, kernel state, nice value, thread count, and start time. Click the row again, the `✕`, or press `Esc` to close.

The table already rendered every field a process carried, but truncated the command and crammed `cwd` inline after an `@`. This lands in two slices: a click-to-open panel over today's fields (phase 1), and five new fields so the panel earns its place beyond "a wider row" (phase 2).

### New per-process fields

| Field | linux (`/proc/<pid>/stat`) | darwin (`ps`) |
| --- | --- | --- |
| `ppid` | field 4 | `ppid=` |
| `state` | field 3 (single char) | first char of `state=` |
| `nice` | field 19 | `nice=` |
| `threads` | field 20 | `null` — no cheap source |
| `startedAtMs` | boot epoch + field 22 ÷ USER_HZ | `null` — no cheap source |

_linux fields are free from the `/proc/<pid>/stat` read the agent already does; the parse is extracted into a pure `parseProcStat` for unit testing. darwin only gains the three standard BSD `ps` keywords — reaching for an unverifiable one (e.g. `etimes`) would risk the whole `ps` invocation failing and dropping every process._ `null` (not `0`) keeps "unavailable" distinct from a real value at the type level.

### Design

Selection is an ephemeral `selectedPid` signal owned by `HostView` (the tier that already owns the process store and filter/sort prefs), reached by `ProcessRow` through a small context — so `ProcessTable` stays a pure renderer with an unchanged signature, and `ProcessDetail` is a `<Show>` sibling (not an overlay), fed the resolved `Process` so the phase-2 fields flowed in with no prop change. Selection clears at the source when its process exits, with a reactive memo as a second guard for the full-snapshot path.

The design plan, including the hickey/lowy review outcomes and the deferred phase 3, is recorded in `docs/plans/process-info-panel.html`.

> Phase 3 — a per-process CPU/mem sparkline — needs per-PID history retention in the parent and is tracked in #21, not part of this PR.

### Try it locally

```sh
nix run github:srid/drishti/feat/process-info-panel
```


srid:

Two parallel reviews on the diff, then a cross-validation pass where each lens audited the other's recommendations.

| 1 | Lowy | `processChanged` silently omits `user` — rendered, filtered, the user-sort comparator; a UID change could freeze the column | Fixed in this PR |
| 2 | Hickey | change-detection was a hand-maintained OR-chain — membership by memory, one edit from a frozen column | Fixed in this PR |
| 3 | Hickey | `threads`/`startedAtMs` used `0` as a darwin "unavailable" sentinel, indistinguishable from a real value | Fixed in this PR |
| 4 | Hickey | `ProcessDetail`'s memory-share re-implemented the `memPct` percent-of-total formula | Fixed in this PR |
| 5 | Hickey ↔ Lowy | include `startedAtMs` in change detection | ⚠️ **No-op** |
| 6 | Hickey ↔ Lowy | panel-scoped (focus) Escape instead of window-global | ⚠️ **No-op** |
| 7 | Hickey | `bootEpochMs = 0` uninitialized sentinel | ⚠️ **No-op** |
| 8 | Hickey | `parsePsLine` uses JS-falsy checks on regex captures | ⚠️ **No-op** |
| 9 | Hickey | "selected pid is live" guarded in two places | ⚠️ **No-op** |

Findings 1 and 2 collapsed into one fix: an explicit `MUTABLE_PROCESS_FIELDS` list (`satisfies (keyof Process)[]`) compared via `.some()` — it includes `user` (discharging Lowy) and makes membership reviewable (discharging Hickey).

### Hickey rationale

- **#2/#1 (Fix)** — the OR-chain is the fragmentation mechanism; it had *already* dropped `user` in this very diff. The field-list makes membership explicit and schema-tied.
- **#3 (Fix)** — `0`-as-unavailable forces every consumer to carry a `> 0` convention out of band; nullable encodes it in the type (`threads` is now `.positive()`).
- **#4 (Fix)** — extract `pctOf(part, whole)` so the guarded share formula lives once.
- **#7 (No-op)** — `bootEpochMs = 0` matches the adjacent `prevWallMs = 0` first-tick idiom in the same closure; boot epoch is never 0. Making only one nullable would diverge from its neighbor.
- **#8 (No-op)** — the regex guarantees non-empty digit captures, so `!ppid`/`!nice` can't misfire on `"0"`; the checks match the existing `!pidStr` idiom.
- **#9 (No-op)** — not redundant: the source-clear handles delta-`removes` + pid recycling; the reactive memo handles a process dropped by a full *snapshot* (which never enters the removes loop). Two paths, two guards.

### Lowy rationale

- **#1 (Fix)** — `processChanged`'s comment promised coverage it didn't deliver; `user` fell in the gap.
- **#5 (No-op, cross-validation)** — `startedAtMs` byte-stability is a property of `proc.ts`'s cached boot epoch, *not* the schema. Including it in `processChanged` would make `main.ts`'s change-detection silently depend on that derivation across a module boundary. The exclusion *is* the decoupling constraint — keep it (the field-list realizes this: `startedAtMs` simply isn't a mutable field).
- **#6 (No-op, cross-validation)** — the window-global Escape keys dismissal off *selection state*, the axis the rest of the feature uses. A focus-scoped handler couples dismissal to DOM focus, which breaks when selection changes via out-of-panel row clicks. The shipped boundary is better.
- **#3 (Fix) is boundary-neutral** — verified the parent bridge passes `Process` objects opaquely, so `null` rides the wire exactly as `0` did.

Fixes landed as commits `ad0f13f` (field-list), `68d6883` (nullable), `c5454eb` (pctOf).

srid:
## Evidence

**Detail panel open** — clicking the `bun --watch src/server/main.ts localhost` row opens the inline panel above the table, showing the full untruncated command, CWD, CPU%, resident memory + % of host RAM, and all five new metadata fields: STATE `sleeping (S)`, PARENT PID, NICE `0`, THREADS `29`, STARTED timestamp. The ✕ close button is visible top-right of the panel.

Process detail panel open

**Process table — default state** — populated with 446 processes sorted by CPU%, no panel open.

Process table default state

srid:

|------|--------|----------|-------------|
| sync | ✓ | 1s | git fetch ok; forge=github |
| research | ✓ | 6m 31s | Mapped agent→parent→client flow; planned 5 new Process fields + context-based selection + sibling panel |
| branch | ✓ | 3s | feat/process-info-panel off origin/master |
| implement | ✓ | 6m 4s | Phases 1+2 across surface/agent/client; parseProcStat + parsePsLine tests; tsc clean |
| check | ✓ | 0s | tsc --noEmit clean |
| docs | ✓ | 20s | README Features + processes-collection row updated |
| fmt | ✓ | 23s | nixpkgs-fmt clean (no .nix changes) |
| commit | ✓ | 20s | 892ac92 pushed |
| hickey+lowy | ✓ | 15m 1s | Ran + cross-validated; 3 Fixed, 5 No-op |
| police | ✓ | 14m 54s | rules + fact-check clean; 2 elegance fixes (0d8ce74) |
| test | ✓ | 20s | bun test 85 pass / 0 fail |
| create-pr | ✓ | 1m 4s | Draft PR + findings ledger |
| ci | ✓ | 2m 20s | justci green, all 16 nodes (linux + darwin), verified vs HEAD |
| evidence | ✓ | 4m 10s | Panel + table screenshots posted |
| **Total** | | **52m 14s** | |

### Optimization suggestions

- **The two review passes dominated (~30m of 52m).** hickey+lowy (15m, incl. a cross-validation round) and police (15m) are agent-driven and inherently slow — but they earned it here: the cross-validation caught a real `user`-omission bug in change detection and rejected two of Hickey's own suggestions. For a diff this size the cost is justified; on a trivial diff, `--minimal` skips both.
- **research took 6m 31s** — most of it mapping the agent-side `/proc` gather path. Pre-reading `packages/app/src/agent/proc.ts` + `main.ts` before invoking `/do` would shave a couple minutes.
- **Both review passes re-edited `proc.ts`**, each triggering a fresh `nix develop` typecheck (~30s of cold-shell startup each). Staying inside one `nix develop` shell for the fix-verify cycles would cut repeated shell-entry overhead.

Workflow completed — PR left as **draft** for review.

PR #25, opened 2026-05-29
Go full-bleed so the process table uses the whole screen width

srid:
**The dashboard now fills the full viewport width instead of sitting in a centered 1152px column.** On a wide monitor the old `max-w-6xl` cap stranded huge margins on the left and right — a process monitor is a density tool, so it should use the glass it's given (the htop instinct).

The wasted space was really *two* caps wearing one coat, and fixing only the obvious one would have made the screen emptier, not denser:

| Cap | Where | Effect |
| --- | --- | --- |
| App shell `mx-auto max-w-6xl` | `App.tsx:348` | centered the whole card, leaving dead margin on wide screens |
| COMMAND cell `max-w-md` | `App.tsx:977` | truncated commands at ~448px — so widening the shell alone would just grow inter-column gaps, not show more |

Both are lifted. The shell goes full-bleed, and the COMMAND column switches to a `w-full max-w-0 truncate` recipe so it *absorbs* the freed width — long `/nix/store/…` paths that used to cut off mid-hash now read in full, with the ellipsis only kicking in on the genuinely longest lines.

> _`w-full` and `max-w-0` are **both** load-bearing in `table-layout: auto`: `w-full` claims the leftover width after the fixed numeric columns, `max-w-0` stops the nowrap content from ballooning the cell past the card's clipped edge (where it would overflow with no ellipsis). Verified by live measurement — dropping `max-w-0` balloons the cell to 2288px and overflows. There's an inline comment so nobody "simplifies" it back into a bug._

### Try it locally

```sh
nix run github:srid/drishti/full-bleed-layout
```

### DIFF
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -345,7 +345,7 @@ export default function App() {
 
   return (
     <div class="min-h-screen bg-gray-50 p-4 font-mono text-sm dark:bg-gray-950">
-      <div class="mx-auto max-w-6xl overflow-hidden rounded border border-gray-300 bg-white shadow-sm dark:border-gray-700 dark:bg-gray-900">
+      <div class="overflow-hidden rounded border border-gray-300 bg-white shadow-sm dark:border-gray-700 dark:bg-gray-900">
         <TabStrip
           hosts={hostList()}
           activeTab={resolvedView()}
@@ -974,7 +974,7 @@ function ProcessRow(props: {
       <td class="px-3 py-0.5 text-right tabular-nums text-gray-700 dark:text-gray-300">
         {formatBytes(rssBytes())}
       </td>
-      <td class="max-w-md truncate px-3 py-0.5 text-left text-gray-700 dark:text-gray-300">
+      <td class="w-full max-w-0 truncate px-3 py-0.5 text-left text-gray-700 dark:text-gray-300">
         <span>{proc()?.command ?? ""}</span>
         <Show when={proc()?.cwd}>
           {(cwd) => (
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -974,6 +974,12 @@ function ProcessRow(props: {
       <td class="px-3 py-0.5 text-right tabular-nums text-gray-700 dark:text-gray-300">
         {formatBytes(rssBytes())}
       </td>
+      {/* COMMAND absorbs the row's residual width and ellipsizes at the cell
+          edge. Both classes are load-bearing in `table-layout: auto`: `w-full`
+          claims the leftover width after the fixed numeric columns, and
+          `max-w-0` stops the `truncate` nowrap content from ballooning the cell
+          past the card (the card clips horizontally, so an un-capped cell would
+          run off-screen with no ellipsis). Dropping either one breaks it. */}
       <td class="w-full max-w-0 truncate px-3 py-0.5 text-left text-gray-700 dark:text-gray-300">
         <span>{proc()?.command ?? ""}</span>
         <Show when={proc()?.cwd}>


srid:

| # | Lens   | Finding                                                            | Disposition    |
|---|--------|-------------------------------------------------------------------|----------------|
| 1 | Hickey | No structural findings — diff resolves a latent coupling          | ⚠️ **No-op**   |
| 2 | Lowy   | Layout-width is a low-volatility axis; no boundary warranted      | ⚠️ **No-op**   |

Both lenses ran on the diff and returned zero "Fix in this PR" findings. Cross-validation was skipped (no applied recommendations to second-guess).

### Hickey rationale

Two declarative class-string edits addressing a single concern (layout width) — no mutable state, signals, conditionals, new files, types, or exported symbols. It introduces no complecting and no concept multiplication. It actually *resolves* fragmentation: the old `max-w-md` cap on the COMMAND cell was implicitly calibrated to the outer `max-w-6xl` cap (two layout decisions coupled by an unenforced convention). Removing the outer cap while keeping `max-w-md` would have broken column stretch; the diff updates both atomically and makes the cell self-describing (`w-full max-w-0 truncate` = "absorb residual width and ellipsize at the boundary") regardless of the outer container. A whole-module read confirmed no straggler element still encodes a stale outer-width cap (`ProcessTable` constrains *height*; the fleet grid uses responsive column counts).

### Lowy rationale

No new file, module, interface, or component split — there is no volatility-boundary placement question to adjudicate, and no prior-encapsulation/duplicate-receptacle trap (there is no existing `LayoutShell` or `--max-app-width` token this parallels). The one identifiable axis — "how wide is the app chrome relative to the viewport" — is low-volatility (one change across the project's lifetime: this PR) and the diff reduces the constraint to an *absence* rather than duplicating it. Extracting a wrapper or token would manufacture a boundary the volatility doesn't pay for. Real but sub-threshold → No-op.

srid:
## Evidence

Full-bleed dashboard on a 1905px-wide viewport — the card spans the whole width (1873px, up from the old 1152px cap), numeric columns stay tight on the left, and the COMMAND column fills the freed space so long `/nix/store/…` paths read in full (ellipsis only on the genuinely longest rows, e.g. the playwright/chrome lines at the far right).

full-bleed process table

Live measurement of the COMMAND `<td>` confirming both width classes are load-bearing:

| Variant | cell width | truncates? | outcome |
| --- | --- | --- | --- |
| `w-full max-w-0` (shipped) | 1578px (fills remaining) | yes, when content exceeds | ✅ correct |
| remove `max-w-0` | 2288px — overflows the 1871px card | no | ❌ clipped, no ellipsis |
| remove `w-full` | 350px | yes | ❌ truncates but wastes width |

srid:

| research | ✓ | 0s | Mapped change: App.tsx:348 app cap + App.tsx:977 COMMAND truncate; no tests assert these classes |
| branch | ✓ | 22s | On feature branch full-bleed-layout |
| implement | ✓ | 8s | Two-line class change: full-bleed card + w-full max-w-0 truncate COMMAND cell |
| check | ✓ | 44s | tsc --noEmit clean |
| docs | ✓ | 0s | No README change: container width isn't documented; detail-panel wording stays accurate |
| fmt | ✓ | 12s | nixpkgs-fmt clean; .tsx untouched (no JS formatter) |
| commit + hickey/lowy | ✓ | 1m 52s | Both lenses No-op (diff resolves latent max-w-md/max-w-6xl coupling; layout-width is sub-threshold for a boundary) |
| police | ✓ | **5m 36s** | All clear after one fix commit (comment-the-non-obvious). Fact-check claim that max-w-0 is a no-op was empirically refuted in a live browser and documented inline. |
| test | ✓ | 26s | bun test: 85 pass, 0 fail. Visual-only diff exempt from coverage; behavior verified live. |
| create-pr | ✓ | 48s | Draft PR #25 + hickey/lowy comment |
| ci | ✓ | 4m 17s | justci: all 16 nodes succeeded; home-manager VM test served HTTP 200; ran against HEAD d7bcbc8 |
| evidence | ✓ | 20s | ## Evidence comment with screenshot + live cell-width measurements |
| **Total** | | **17m 4s** | |

**Slowest step**: `police` (5m 36s)

### Optimization suggestions

- **`police` dominated (5m 36s, ~33%)** — almost entirely the live-browser fact-check: spinning up a dev server + chrome-devtools to disprove the "max-w-0 is a no-op" claim. That verification was worth it (it caught a wrong finding and forced an explanatory comment), but it's the kind of thing that only pays off once — a re-run of this branch via `--from ci-only` skips it.
- **`ci` was 4m 17s** — the cost is the home-manager NixOS VM test, not the diff. For iterative pushes use `--from ci-only` to resume just CI instead of replaying research→police.
- **This was a 4-line CSS diff** — `--minimal` would have skipped docs/hickey+lowy/police/evidence and shaved ~13m. It would also have missed the `comment-the-non-obvious` finding, so the trade was real here; worth `--minimal` only when the diff is genuinely mechanical.

Workflow completed.

PR #26, opened 2026-05-29
Show root-filesystem disk usage alongside CPU and memory

srid:
**Each host now reports disk fullness as a third headline metric, right next to CPU and memory.** drishti already streamed CPU% and memory% from every host; this adds root-filesystem (`/`) %full — a fleet-card row, a header stat, and a third trace in the 30m history sparkline. It's a slow-moving *capacity gauge* (df-style "how full is the disk"), not disk I/O.

Collection deliberately mirrors the memory path: a scalar `diskUsed`/`diskTotal` on the `system` cell, so it folds into the existing scalar history ring for free. The agent reads it via the `statfs("/")` syscall — the one capacity source universal across Linux and macOS (no `/proc` file exposes free space, and `statfs` is unprivileged on both).

```
statfs("/")  ─▶  diskUsed/diskTotal  ─▶  system cell  ─▶  history ring  ─▶  sparkline
  (agent)         (SystemSchema)         (parent samples)     (cpu/mem/disk traces)
```

### What you see

- **Fleet card** — a `disk` usage bar (`/ · 188/250 GB · 75%`) under the cpu/mem bars, plus a third (amber) trace on the 30m sparkline.
- **Per-host view** — a `disk` stat in the header and a third trace + legend chip in the history chart, across every 1m/5m/15m/30m window.

### One refactor rode along

The client's per-series render sites (sparkline polylines, both legends, the projection) were hardcoded twice over for cpu/mem — adding a third would have meant editing each. A single `SERIES` descriptor table now drives them, so disk is **one entry**. It's keyed by `MetricKey` and pinned with `satisfies Record<MetricKey, …>`, so the compiler errors if the render set and the data set ever drift (a missing series can't silently render a blank trace).

> **Mount-selection policy: root `/` only.** Unlike memory, a host has many filesystems and no single capacity aggregate. This scalar reports `/` by design — a host that splits `/var`, `/nix`, or `/data` onto separate disks won't see those here. A per-mount view would be a future `diskDevices` collection (mirroring `networkInterfaces`), not a reinterpretation of this field. The choice is documented at the schema and the reader.

### Try it locally

```sh
nix run github:srid/drishti/feat/disk-usage
```


srid:

| 1 | Hickey | `SERIES` ↔ `MetricKey` kept in sync by unenforced convention (missing entry → silent blank trace; unsafe `as Record<MetricKey,string>` cast) | Fixed in this PR |
| — | Lowy | No findings — addition lands cleanly on existing volatility boundaries | No-op |
| 2 | Lowy (cross-validate) | Audited Hickey's `satisfies Record<MetricKey,…>` fix — confirmed it tightens the data/render seam without misplacing a boundary | Fixed in this PR |

### Hickey rationale

The diff adds disk as a third headline metric and is structurally clean: each touched module carries one concern (collection → wire schema → derivation → history ring → render), no cycles, no new top-level abstractions. `diskPct`/`diskGb` extend `memPct`/`memGb` rather than parallel them, and the new `SERIES` table is a *net concept reduction* — it single-sources render metadata that was previously hardcoded across the sparkline, both legends, and the projection.

One finding: the "which metrics exist" set was split between `MetricKey` (mechanically enforced via the union, `MetricSample`, `captureSample`) and the `SERIES` render table, typed `ReadonlyArray<{ key: MetricKey; … }>`. That array type catches a *stale* key but permits a *missing* one — a metric added to `MetricKey` without a `SERIES` entry would compile, transmit, sample, and project, then silently render nothing. `projectHistory`'s `as Record<MetricKey, string>` cast asserted an exhaustiveness the data didn't guarantee. **Fixed** (`7d48cca`) by keying the metadata as a `Record<MetricKey, SeriesMeta>` pinned with `satisfies`, deriving the ordered iteration array from it — divergence in either direction is now a compile error, and the cast is provably safe.

Cleared as no-op: `diskBytesFromStatfs` (pure, isolated), `readRootDiskUsage` (atomic both-zero-on-failure invariant), the `...disk` spread in `readSystem` (flat composition, not a complect), `diskGb` zero-guard.

### Lowy rationale

No actions. The decomposition tracks volatility, not just functionality:

- **Metric source** (per-platform read) stays in `createProcReader` + the platform readers — `readRootDiskUsage` (`statfs`) sits there correctly.
- **Metric set** expands across its two deliberate receptacles: the data layer (`MetricKey`/`MetricSample`/`captureSample`) and the render layer (`SERIES`). Each touch is a single well-named site, not a fan-out — and `captureSample`'s comment names the design.
- **`SERIES` encapsulates a previously-sprayed volatility** (color/label pairs repeated across four hand-coded sites) — a genuine new receptacle, not a parallel duplicate.
- **`diskPct`/`diskGb` reuse the existing "scalar capacity metric" receptacle**, symmetric with `memPct`/`memGb`.
- **The future `diskDevices` collection is correctly deferred, not pre-built** — the scalar is the right receptacle for the current root-only policy, and the future seam is named at the schema.

Cross-validation: Lowy audited Hickey's `satisfies` recommendation and confirmed it keeps the data-layer/render-layer boundary in place — `MetricKey` is the stable interface the render layer must already enumerate, so keying by it is structural necessity, not accidental coupling.

srid:
## Evidence

**Fleet overview** — host card showing the new `DISK` usage bar (`/ · 684.2/973.0 GB · 70%`) beneath the existing CPU and MEM bars, plus the 30-minute sparkline with three colour-coded traces (emerald = cpu, indigo = mem, amber = disk) and the `disk` legend chip:

Fleet overview with disk bar and 3-trace sparkline

**Per-host view** — header stats row with the new `DISK … GB (…%)` stat alongside cpu/mem, and the history chart with the amber disk trace and `disk` legend chip showing the live percentage:

Per-host view with disk header stat and amber history trace

srid:

| research | ✓ | — | Confirmed metric flow; `fs.statfs` works in Bun; scoped 6-file change + SERIES-table refactor |
| branch | ✓ | 4s | feat/disk-usage from origin/master |
| implement | ✓ | — | Disk %full end-to-end (statfs reader → SystemSchema → MetricSample → diskPct/diskGb → captureSample → fleet card + header + sparkline); SERIES table de-fragments render sites; tests added |
| check | ✓ | — | `tsc --noEmit` clean |
| docs | ✓ | 41s | README updated (tagline, fleet-card list, feature bullet, surface table) |
| fmt | ✓ | 21s | nixpkgs-fmt clean (no JS formatter configured) |
| commit | ✓ | 14s | `8a3814d` pushed (10 files) |
| hickey+lowy | ✓ | — | hickey: 1 finding → fixed `7d48cca`; lowy: clean; cross-validated |
| police | ✓ | 4m 45s | 3 passes clean; 1 elegance fix committed |
| test | ✓ | 18s | 58 pass / 0 fail; new behavior covered |
| create-pr | ✓ | — | Draft PR + hickey/lowy ledger |
| **ci** | ✓ | **17m 49s** | all 16 justci nodes green; GH rollup success for `5652eac` |
| evidence | ✓ | — | Fleet + per-host screenshots posted |
| **Total** | | **44m 11s** | |

### Optimization suggestions

- **`ci` dominated (17m 49s)** — most of it the `home-manager` NixOS VM build, plus a wasted first attempt that bailed on a dirty tree (the untracked talk-mode `.html` artifact) and left `home-manager@x86_64-linux` orphaned-pending. For re-runs use `/do --from ci-only`; and keep scratch artifacts out of `$PWD` (or `.gitignore` them) so justci's strict-mode HEAD pin doesn't refuse the first run.
- **`police` was 4m 45s** — the elegance/`simplify` pass is the cost; on a diff this size it found only one comment-level fix, so `--minimal` would have been a reasonable call if speed mattered.
- **Pre-warm the dev shell** — `check` had to `bun install` a missing dep (`cleye`) before `tsc` could run; running `bun install` once up front shaves that off the critical path.

Workflow completed.

PR #27, opened 2026-05-30
Distinguish terminal connection failure from transient retry

srid:
**A host that drishti has permanently given up on now says so** — instead of showing "Disconnected. Retrying…" forever while the parent's reconnect loop has long since stopped. The real, usually-fixable error (an ssh-agent miss, or a remote `nix-daemon` that won't accept the closure because your user isn't in `trusted-users`) used to live only in the parent's stderr; it now reaches the browser, alongside a **Reconnect** button that re-arms the session in place.

The screen was conflating two independent loops. The browser↔parent WebSocket reconnects forever; the parent↔agent ssh/`nix copy` link retries 5× and then **gives up**. Both rendered as one red "Retrying…", so a terminal failure was indistinguishable from a transient blip.

### The fix

The root cause was an impoverished wire contract: the `connection` cell was a flat 4-value enum with no room for "why" or "is this terminal". It now mirrors the `HostSession` library's richer state:

| state | before | after |
| --- | --- | --- |
| `disconnected` | red, "Disconnected. Retrying…", no pulse | amber, **"Reconnecting…"**, pulsing — honest about the transient gap |
| `failed` | *(did not exist — folded into `disconnected`)* | red terminal card: the real `lastError`, a remediation hint, and a **Reconnect** button |

`lastError` and `progressLines` (the live "attempt N/5" tail) are now forwarded at the single `onState` mirror in `server/router.ts` instead of being dropped. Recovery rides a new `hosts.reconnect` admin RPC — deliberately *not* host add/remove (that races the fleet-view fallback and churns the host set); re-arming a session is a distinct concern from host-set membership.

### Cross-repo

The terminal `failed` state and the `HostSession.reconnect()` method live in the library, which is the layer that owns the reconnect loop. This PR pins that companion change:

> Depends on **juspay/kolu#1039** — pinned via `npins/sources.json` to the PR branch. Re-pin to `master` once that merges.

The design rationale (and the reviewer pass that shaped it) is captured in `docs/plans/talk-reconnect-ux.html`.

### Try it locally

```sh
nix run github:srid/drishti/reconnect-failed-state
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ just regenerate-bun-nix           # after any bun.lock change
 
 `just dev` exports `DRISHTI_AGENT_DRVS_JSON` (the per-system `.drv` map from the flake's `agentDrvsJson` attribute) and boots Bun in watch mode. The parent probes each host's nix-system on add (asking the host's own Nix for `builtins.currentSystem`) and picks the matching `.drv` from the map — so one dev session can mix architectures. The dev server invokes `buildClient()` at startup so a single `bun --watch` covers both server-TS and client-bundle rebuilds. Browser refresh is manual — there's no HMR.
 
-The first connect to a fresh remote ships the agent closure over `ssh` (`nix copy --derivation` then `nix-store --realise`); subsequent connects reuse it. The progress is streamed to the browser via the `connection` cell.
+The first connect to a fresh remote ships the agent closure over `ssh` (`nix copy --derivation` then `nix-store --realise`); subsequent connects reuse it. The progress is streamed to the browser via the `connection` cell. A dropped link reconnects automatically (the tab pulses amber, "Reconnecting…"); if it keeps failing — e.g. the remote `nix-daemon` won't accept the closure because your user isn't in `trusted-users` — the parent eventually gives up and the host enters a terminal **failed** state. drishti then shows the underlying error and a **Reconnect** button that re-arms the session in place, so you don't have to restart the parent.
 
 ## Deployment (home-manager)
 
--- a/docs/plans/talk-reconnect-ux.html
+++ b/docs/plans/talk-reconnect-ux.html
@@ -1,0 +1,261 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>Reconnect UX — design</title>
+<style>
+  :root {
+    --bg: #0b1220; --panel: #111a2b; --panel2: #0e1726; --border: #1e2a3f;
+    --fg: #e5e9f0; --muted: #8b97ad; --dim: #5d6b82;
+    --emerald: #10b981; --amber: #f59e0b; --red: #ef4444;
+    --mono: ui-monospace, SFMono-Regular, "SF Mono", Menlo, Consolas, monospace;
+  }
+  * { box-sizing: border-box; }
+  body {
+    margin: 0; background: var(--bg); color: var(--fg);
+    font: 14px/1.6 var(--mono); padding: 32px 28px 80px;
+  }
+  .wrap { max-width: 920px; margin: 0 auto; }
+  h1 { font-size: 20px; margin: 0 0 4px; letter-spacing: .3px; }
+  h2 { font-size: 15px; margin: 34px 0 10px; color: var(--fg);
+       border-bottom: 1px solid var(--border); padding-bottom: 6px; }
+  h3 { font-size: 13px; margin: 20px 0 8px; color: var(--muted); font-weight: 600; }
+  p { margin: 8px 0; color: var(--fg); }
+  .muted { color: var(--muted); } .dim { color: var(--dim); }
+  code { background: var(--panel2); padding: 1px 5px; border-radius: 4px;
+         border: 1px solid var(--border); font-size: 12.5px; }
+  a.cite { color: var(--emerald); text-decoration: none; }
+  .repo { display:inline-block; font-size: 10.5px; padding: 0 6px; border-radius: 4px;
+          border: 1px solid var(--border); color: var(--muted); margin-right: 4px; }
+  .repo.kolu { border-color:#c792ea; color:#c792ea; } .repo.dr { border-color:var(--emerald); color:var(--emerald); }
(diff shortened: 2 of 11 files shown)


srid:

| 1 | Hickey | Flat `ConnectionInfo` permits type-level illegal states (`lastError` nullable on any `state`) | ⚠️ **No-op** |
| 2 | Lowy | `progressLines` has no length cap on the wire | Fixed in this PR |
| 3 | Hickey (cross-val) | `.max(20)` would couple drishti to kolu's *private* `MAX_PROGRESS_LINES` — document the absence instead | Fixed in this PR |
| 4 | Lowy (cross-val) | Flat mirror (not a discriminated union) is the right encapsulation of link-state volatility | ⚠️ **No-op** |

### Hickey rationale
No complecting or fragmentation to fix. The survey for prior patterns came back clean: `FailedCard` is a genuinely new surface (the only prior error UI is `AddHostForm`'s one-line inline validation — a different *kind* of thing), the overlay's `<Show>` branch is idiomatic Solid, and `hosts.reconnect` extends the existing admin-RPC shape rather than reinventing it.

**F1 (No-op):** the flat `ConnectionInfo` lets `connected` carry a non-null `lastError` at the type level. A discriminated union would forbid that — but it would re-model kolu's lifecycle vocabulary inside drishti, so a future kolu state addition would force a union change here where the flat enum needs only a one-line bump. Lower coupling to the tier that owns the lifecycle wins; consumption is guarded (`lastError` read behind `state === "failed"`).

### Lowy rationale
The `connection` cell is the correct, sole receptacle for link-lifecycle volatility — the diff *expands* it rather than spawning a parallel diagnostics channel. Forwarding `lastError`/`progressLines` does **not** leak the library's vocabulary: `ConnectionInfo` is defined independently and the `onState` mirror is the single coupling point. `hosts.reconnect` is correctly a distinct volatility from add/remove (session lifecycle vs host-set membership — different persistence, different blast radius).

**Finding #2:** Lowy flagged that `progressLines` had no wire-level cap while the library trims to 20. The first-pass disposition was "add `.max(20)`".

### Cross-validation
Both lenses produced findings, so each re-ran against the other's output:

- **Hickey vs. Lowy's `.max(20)` (#3):** reversed it. `MAX_PROGRESS_LINES` is *private* to `HostSession` (not exported), so baking `20` into drishti's parse-time schema would couple the wire contract to a constant drishti doesn't own — and silently reject valid frames if kolu ever raised the cap. Since the UI reads only `progressLines.at(-1)`, the bound is behaviorally irrelevant. Resolution: document the deliberate absence (commit `7b9d9cd`) rather than enforce a foreign constant.
- **Lowy vs. Hickey's F1 (#4):** confirmed No-op. Under the volatility lens the flat shape yields a *more* stable interface — the `router.ts` mapping can absorb a future kolu union without touching browser consumers; a surface-level union would couple the browser's match arms to kolu's internal state partitioning.

srid:
## Evidence

**Connected — normal htop view** (proves the widened `connection` cell doesn't break connected rendering):

Connected — process table fully populated, green dot on localhost tab

**Reconnecting — amber "Reconnecting…" overlay** (staged by killing the agent process mid-session; server logs show `connection: connected → disconnected` then `reconnecting in 8000ms… (attempt 3/5)`):

Reconnecting — pulsing amber overlay with progress line

**Failed — red error card with ↻ Reconnect button** (reached after exhausting all 5 retry attempts; server emits `connection: disconnected → failed`):

Failed — red card showing error, trusted-users hint, and Reconnect button

> The `failed`/`reconnecting` states were staged by `SIGKILL`-ing the agent subprocess from the same machine. The add-host dialog path can't reach them because the arch-probe (`builtins.currentSystem` over ssh) gates non-localhost hosts and rejects unreachable targets before a session is created.

srid:

| sync | ✓ | 0s | forge=github |
| research | ✓ | 1m 15s | file:line map for both repos |
| branch | ✓ | 10s | reconnect-failed-state from origin/master |
| implement | ✓ | 10m 28s | surface widened; router forwards lastError/progressLines; disconnected/failed split; FailedCard + Reconnect; hosts.reconnect RPC; npins → kolu 39b0cb14 |
| check | ✓ | 23s | just typecheck EXIT=0 (kolu reconnect() resolved) |
| docs | ✓ | 27s | README connection section updated |
| fmt | ✓ | 27s | just fmt clean |
| commit | ✓ | 0s | feature + plan artifact pushed |
| hickey+lowy | ✓ | 0s | Hickey clean; Lowy .max(20) reversed by cross-validation → doc fix |
| police | ✓ | 7m 6s | rules+fact-check clean; 1 elegance reuse fix |
| test | ✓ | 28s | bun test 108 pass / 0 fail |
| create-pr | ✓ | 1m 3s | PR #27 + hickey/lowy comment |
| ci | ✓ | 2m 28s | all 16 nodes green (linux+darwin); covers HEAD |
| evidence | ✓ | 5m 35s | live connected/reconnecting/failed screenshots |
| **Total** | | **40m 53s** | |

**Slowest step**: `implement` (10m 28s)

### Optimization suggestions

- **`implement` (10m 28s) dominated** because it spanned two repos in one pass — kolu impl + push + PR, npins bump, then the drishti change. That was the right call (the pin stays coherent in one run), but for re-runs `--from ci-only` skips it entirely.
- **`police` (7m 6s) was second** — it spins up its own background CI. When iterating, running `/simplify` *before* `/do` pre-empts the elegance pass (here it found one reuse fix).
- **`evidence` (5m 35s):** the `failed`/`reconnecting` states can't be staged via the add-host dialog (the arch-probe gates non-localhost adds) — the working recipe was `SIGKILL`-ing the agent subprocess. Worth adding to `.agency/do.md`'s evidence section so the subagent doesn't rediscover it.

> **Cross-repo:** depends on [juspay/kolu#1039](https://github.com/juspay/kolu/pull/1039) (pinned via `npins`). kolu CI was run separately: the change's own lanes (typecheck-darwin, biome, nix-build, e2e-linux, unit-linux) are green; the reds are a `tsc`-not-on-PATH env flake in `packages/shared` on the linux lane and a pre-existing flaky iframe-preview e2e on darwin — neither touches `surface-nix-host`. Re-pin to `master` once #1039 merges.

Workflow completed.

PR #28, opened 2026-05-30
Fix MEM column wrapping and USER column over-widening in process table

srid:
**The process table's MEM value no longer wraps onto two lines, and the USER column no longer hogs horizontal space.** Both were symptoms of how the auto-layout table distributed its slack width: COMMAND uses `w-full max-w-0` to absorb leftover width and ellipsize, but `max-w-0` caps how far COMMAND can grow — so the table's residual width spilled into the other auto-sized columns and inflated USER, while the un-`nowrap`'d MEM cell let `"817.5 MB"` break across two rows.

The fix gives the four leading cells (PID / USER / CPU% / MEM) `w-px whitespace-nowrap`, collapsing each to its content width and leaving **COMMAND as the sole absorber of the row's slack**.

| Symptom | Cause | Fix |
| --- | --- | --- |
| MEM reads `817.5` / `MB` on two lines | cell had no `whitespace-nowrap`, wrapped at the space | `whitespace-nowrap` keeps it on one line |
| USER column stretched far past its content | COMMAND's `max-w-0` capped its growth, so slack inflated other auto columns | `w-px` collapses the fixed columns to content width |

_Pure CSS/className change — no behavioral logic touched; `tsc` and the client test suite (51 tests) pass._

### DIFF
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -1115,14 +1115,25 @@ function ProcessRow(props: {
           : "hover:bg-gray-50 dark:hover:bg-gray-800/40"
       }`}
     >
-      <td class="px-3 py-0.5 text-right tabular-nums">{props.pid}</td>
-      <td class="px-3 py-0.5 text-left">{proc()?.user ?? ""}</td>
+      {/* The four leading columns shrink to their content via `w-px` +
+          `whitespace-nowrap`: `w-px` makes each column's preferred width
+          tiny so the auto-layout table collapses it to its (nowrap) content
+          width, leaving COMMAND's `w-full` as the sole absorber of the row's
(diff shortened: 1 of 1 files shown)


srid:
## Evidence

Screenshots captured from the live drishti dev server running on **vanjaram** (x86_64-linux NixOS, 32-core, 134.8 GB RAM), monitoring vanjaram's own 567 processes via the fix branch `52134bc`.

**Default view (CPU-sorted):**

process table

**Sorted by MEM descending** (top entries: 565.9 MB, 206.9 MB, 203.6 MB, 188.0 MB, …):

process table sorted by MEM

All MEM values (e.g. "565.9 MB", "206.9 MB") render on a single line, and the USER column is snug to its content — COMMAND is the sole flex-grow column absorbing all remaining width.

PR #29, opened 2026-05-30
Time out hung connects so a host can't sit in "Connecting" forever

srid:
**A host could wedge in "Connecting…" forever.** When an ssh link dropped and the parent respawned the agent, the transport would sometimes come back up — the remote logged `serving surface over stdio` — but the first RPC never roundtripped and the agent process never exited. The parent's `HostSession` armed *no* timeout on the `connecting` phase, so nothing transitioned the state: no error, no reconnect, no give-up. Just a spinner, indefinitely.

The fix is a **connect-phase watchdog** in the kolu library that owns the link lifecycle, plus a small bit of in-overlay feedback here.

### The two exits from `connecting` weren't enough

```
copying ──provision ok──▶ connecting ──first RPC──▶ connected
                          connecting ──child exit──▶ disconnected ──retry──▶ failed
                          connecting ──(nothing)───▶ 🪦 hung forever
```

`markConnected()` (first RPC) and the child `exit`/`error` handlers were the only ways out. A transport that's *up* but whose handshake wedges — child alive, no exit — hit none of them. The watchdog adds the missing edge: armed when the child spawns, it kills a wedged connect after `connectTimeoutMs` (default 30s), which routes through the ordinary exit handler into the **existing** retry-then-`failed` machinery. So a hung connect now reaches the terminal `failed` card and its Reconnect button like any other failure, instead of hanging invisibly.

> The watchdog itself lives in **[juspay/kolu#1048](https://github.com/juspay/kolu/pull/1048)** (merged); this PR repins npins to pick it up. The library change also folds the reconnect-backoff timer and the new watchdog into one `pendingTimer` (they're never live at once), making "at most one pending timer" a structural invariant.

### Elapsed-time feedback while connecting

The overlay now shows how long the current phase has been running — **"Connecting… 18s"** — reset on each state change, so a slow connect reads as abnormal *before* the watchdog trips. The suffix is a pure `withElapsed()` helper (unit-tested); the ticker only runs in in-progress states.

### Review refinements

| Lens | What landed |
| --- | --- |
| Hickey | Gate the elapsed ticker on the canonical `STATE[state].pending` flag, not on which branch happens to read the signal |
| Lowy | Pass `connectTimeoutMs` explicitly at the `getHostSession` call as a named const tied to `wire.ts`'s 60s socket deadline — the 30s budget must stay under it, so state it where the constraint lives |

The design rationale and the Hickey/Lowy rehearsal are captured in `docs/plans/talk-stuck-connecting.html`.

### Try it locally

```sh
nix run github:srid/drishti/connect-watchdog
```


srid:

Reviewed the full cross-repo diff (kolu watchdog + drishti overlay), first-pass in parallel then a cross-validation pass where each lens audited the other's recommendations.

| 1 | Hickey | Elapsed ticker re-armed in `connected`/`failed` where the counter isn't rendered — gate on the canonical `STATE[state].pending` flag | Fixed in this PR (`8a29b6c`) |
| 2 | Lowy | drishti relied on kolu's 30s default implicitly — the budget is meaningful only relative to `wire.ts`'s 60s socket deadline; state it explicitly at the call site | Fixed in this PR (`c40968d`) |
| 3 | Hickey | `armTimer` precondition ("callers arm only when `pendingTimer` is null") is documented, not enforced — add a defensive `clearTimeout`+log | ⚠️ **No-op** |
| 4 | Hickey | spawn-scoped `connectTimedOut` flag splits "why did the child exit" across writer/reader | ⚠️ **No-op** |
| 5 | Hickey | `pendingTimer` fold (reconnect timer + watchdog → one slot) | ⚠️ **No-op** (fragmentation *repair*) |
| 6 | Lowy | Watchdog placement inside `HostSession` (owns link-lifecycle volatility) | ⚠️ **No-op** (correct seam) |
| 7 | Both | `withElapsed` vs existing `formatUptime` — concept multiplication? | ⚠️ **No-op** (distinct axes) |

Finding #3 was Fix-in-this-PR on Hickey's first pass, then **flipped to No-op by Lowy's cross-validation**: the single-slot invariant is already structurally enforced by the `updateState` choke-point + the `scheduleReconnect`/`teardown` guards, and a defensive clear inside `armTimer` would add a *second* enforcement point at a different layer and soften a "can't happen" into a runtime warning. Left as shipped.

### Hickey rationale

The substantive kolu rework is a net simplicity improvement — it *repairs* fragmentation rather than introducing it. Folding `reconnectTimer` and the watchdog into one `pendingTimer` (#5) enforces "at most one pending phase-transition timer" structurally instead of leaving two parallel slots the domain says must never both be live; the single `updateState` choke-point (clear on leaving `connecting`) eliminates the scattered "remember to clear X" discipline. The `connectTimedOut` closure flag (#4) is mechanically safe under the single-threaded event loop and the de-coupled alternative duplicates the entire exit-handler body — a worse trade. `withElapsed` (#7) *annotates* a status string with exact seconds (0–30s diagnostic) while `formatUptime` *coarsens* a large value to minutes — different domains, output conventions, and shapes; placement in `connectionColors.ts` is correct. Only actionable bite was #1 (ticker liveness should consult the `pending` axis, not which branch reads the signal).

### Lowy rationale

The three core boundary placements are correct. The watchdog (#6) is the same link-lifecycle volatility axis `HostSession` already encapsulates (reconnect loop, backoff, give-up gate) — a new edge case in "how `connecting` terminates", not a new axis. `connectTimeoutMs` is correctly surfaced as a `HostSessionOptions` tunable, sibling of `reconnectDelayMs`. The elapsed-overlay ticker is a presentation projection (state-change event → wall-clock elapsed) that belongs at the view layer. The one real finding (#2) was a call-site interface-stability leak: drishti owned a constraint (server must give up before the 60s browser socket does) that lived only in the library's doc comment — now an explicit named const at the call site. On cross-validation, Lowy agreed with Hickey #1 for a boundary reason too: reading `STATE[state].pending` for ticker control routes the question to the encapsulated classification, the right seam.

srid:
## Evidence

The **connected** frame is a live end-to-end capture — real agent data flowing agent → parent → browser. The three overlay frames are the **real** overlay component rendered with representative connection data via a temporary dev-only `?_evi=<state>` override, since a wedged/failed remote isn't reproducible on a single-host dev box (the connection cell only reaches `connecting`/`disconnected`/`failed` after a host provisions and *then* stalls at the handshake — the exact bug this PR fixes).

### Connected — live process table
connected
~400 PID rows, full CPU/MEM/DISK strip and load averages — the pipeline is alive end-to-end.

### Connecting — live elapsed-time suffix
connecting
The new `withElapsed()` suffix ticks up live — here "Connecting… 9s" — so a slow connect reads as abnormal before the 30s watchdog trips it.

### Reconnecting — transient amber retry
reconnecting
The transient `disconnected` state is now its own amber "Reconnecting… Ns" with the parent's live progress line ("reconnecting in 4000ms… (attempt 2/5)") — no longer conflated with give-up.

### Failed — terminal card with error, hint, and Reconnect
failed
The terminal `failed` state (where a hung connect now lands, instead of spinning forever) renders the red "Couldn't reach this host" card: verbatim `lastError`, the `trusted-users` remediation hint, and a ↻ Reconnect button wired to `hosts.reconnect`.

PR #31, opened 2026-05-30
Make Reconnect actually re-arm a failed host

srid:
**The "Reconnect" button on a failed host did nothing** — no state change, no console error, nothing. It only broke for the most common real-world failure: when the link gave up because `nix copy --derivation` couldn't push the agent closure (an untrusted-user / unsigned-closure remote). This makes the button work, and replaces a misleading one-size-fits-all hint with the host's actual error.

### Root cause (in the `@kolu/surface-nix-host` library)

`HostSession.reconnect()` guards on `clientPromise !== null` to avoid stacking spawns. That slot is nulled only when an ssh child exits — but a `nix copy` failure throws out of `spawn()` *before any child is created*, and the give-up path transitioned to the terminal `failed` state **without** clearing it. So the slot kept a stale rejected promise, `reconnect()` saw "spawn already in flight" and returned immediately. The admin RPC still resolved `{ ok: true }`, so nothing surfaced anywhere.

Fixed upstream in **juspay/kolu#1053** (clears `clientPromise` on the `failed` transition, via a named `clearClientPromise()` all three terminal paths route through, with the package's first vitest regression test). This PR consumes it via the npins pin.

### What landed here

- **Reconnect works** — npins pin bumped to the kolu fix.
- **Real error, not a guess** — the failed-host card dropped its hardcoded _"your user probably isn't in `trusted-users`"_ tip (shown for *every* failure regardless of cause) and now renders the captured connection log tail — the actual `nix copy`/`ssh` output. `lastError` stays as the terse headline; the log carries the why.
- **`.agency/code-policy.md`** — codifies _"errors must reach the user, never the void"_, the rule this bug broke three ways (silent no-op, success-on-failure RPC, speculative tip decoupled from the real error).

### Structural review fixes (hickey + lowy, then police)

| Finding | Resolution |
| --- | --- |
| `logTail` recomputed per access (hickey) | `createMemo`, matching the file's reactivity idiom |
| `surface.ts` comment claimed "UI reads only the last line" (lowy) | corrected — the card reads the last few |
| `reconnect` RPC swallowed a throw into ignored `{ ok: false }` (lowy → police) | reverted to the plain sync handler so any improbable throw propagates to the client's `.catch` instead of vanishing |

> **Cross-repo dependency — resolved.** juspay/kolu#1053 is merged; the npins pin now tracks **kolu master** (`9f2a9e6a`), so `sources.json`'s `branch: master` and its revision agree. The bump also pulls kolu's rightPanel resize-coalesce (#1041), which added a `@solid-primitives/scheduled` dep to `@kolu/surface` — mirrored into drishti's deps (bun.lock/bun.nix regenerated).

### Try it locally

```sh
nix run github:srid/drishti/fix/reconnect-noop-error-visibility
```


srid:

Run post-implement on the real two-repo diff (drishti + the kolu fix). Cross-validation (each lens auditing the other's recommendations) ran and was clean.

| 1 | Hickey | `logTail` recomputed on each access | Fixed in this PR |
| 2 | Hickey | `clearClientPromise` double-duty (mutex + client handle) not split | ⚠️ **No-op** — correct to leave; splitting is scope creep, fix made `null` honest |
| 3 | Lowy | `surface.ts` comment "UI reads only the last line" stale | Fixed in this PR |
| 4 | Lowy | npins pin tracks an unmerged kolu PR branch | Fixed in this PR (re-pin follow-up hoisted into PR body) |
| 5 | Lowy | `reconnect` RPC could swallow a throw as `{ ok: true }` | Fixed, then **superseded by police** — sync handler propagates instead |
| 6 | Lowy | root-cause fix belongs in kolu; classify-nothing log display | ⚠️ **No-op** — boundary confirmed correct |

### Hickey rationale

The `clearClientPromise()` extraction is a genuine convention→structure move: the "no spawn in flight ⇒ `clientPromise` is null" invariant now has one named choke-point instead of three scattered `= null` assignments one of which (the give-up branch) forgot it. The FailedCard change *removes* complecting — a one-cause-fits-all tip replaced by the real log tail (headline `lastError` + context log are distinct projections, not redundant). Only concrete finding: `logTail` was a plain arrow read twice in JSX → `createMemo`, matching the file's own reactivity idiom. The test's "copying set synchronously" assertion couples to `spawn()`'s pre-await ordering — acceptable for a regression pin, flagged as the most likely false-alarm on unrelated refactors.

### Lowy rationale

Boundary is correct: the link-lifecycle volatility (state machine, retry gate, `clientPromise` invariant) stays encapsulated in kolu's `HostSession`; drishti only bumps the pin and delegates — no local workaround that would leak lifecycle knowledge across the boundary. Rendering the raw `progressLines` log is *not* a coupling risk because the display never branches on its content (`slice(-8).join`), honoring the cell contract's "display text, never a branch condition". Differentiated remediation tips, if ever wanted, belong behind a structured `failureCause` discriminant minted in kolu — a correctly *deferred* seam, not a missing one. The npins-tracks-a-PR-branch smell is real and mitigated by the re-pin follow-up now in the PR body.

> Finding #5 is where the reviewers interacted: Lowy suggested mirroring the sibling `remove` handler's try/catch; the subsequent **code-police** pass showed that — because the client consumer only has a `.catch` and ignores a resolved `{ ok: false }` — wrapping a (practically impossible) throw into `{ ok: false }` is *strictly worse* for visibility than letting it reject. The plain sync handler won.

srid:

| branch | ✓ | 13s | fix/reconnect-noop-error-visibility off origin/master |
| implement | ✓ | 7m 57s | kolu fix+test+PR #1053 (aa64491e); drishti npins bump, FailedCard shows progressLines, .agency/code-policy.md |
| check | ✓ | 1m 12s | bun typecheck exit 0 against rebuilt kolu pin |
| docs | ✓ | 22s | README failed-state paragraph updated |
| fmt | ✓ | 22s | just fmt clean; TS matches house style |
| commit | ✓ | 33s | de81275 pushed |
| hickey+lowy | ✓ | 7m 32s | both-repo diff; 3 findings fixed; cross-validation clean |
| police | ✓ | 6m 54s | rules+fact-check clean; elegance dropped dead try/catch (cfa8fad) |
| test | ✓ | 15s | bun test 114 pass / 0 fail; reconnect covered by kolu vitest |
| create-pr | ✓ | 1m 9s | draft PR #31 + hickey/lowy comment |
| ci | ✓ | 2m 49s | justci: all 16 nodes succeeded against HEAD cfa8fad |
| evidence | — | 1m 22s | judgment-skipped: terminal `failed` card needs a remote that passes the ssh arch-probe but fails `nix copy` — not reproducible in the local sandbox; behavior proven by the kolu regression test (red→green) + justci |
| **Total** | | **32m 40s** | |

**Slowest step**: `implement` (7m 57s)

### Optimization suggestions

- **`implement` dominated (7m 57s)** — mostly cross-repo setup: a fresh kolu git worktree + `pnpm install` + nix-devshell warmup to stand up vitest in `surface-nix-host` (which had no test runner). Pre-warming the kolu worktree's devshell before `/do` would shave minutes.
- **Structural review was the real wall-clock** — `hickey+lowy` (7m 32s) + `police` (6m 54s) ≈ 14m, nearly half the run. That's fixed overhead independent of diff size; on a genuinely small/confined change `--minimal` skips it. This diff was cross-cutting (two repos, a new policy doc), so the full pass earned its keep — police even caught that a lowy fix regressed error-visibility.
- **Cross-repo ordering paid off** — pushing the kolu branch *before* the npins bump let drishti CI build against the real fix in one green pass, rather than discovering an unbuildable pin late.

> **Merge-blocker:** the npins pin tracks the kolu PR branch (`aa64491e`). Re-pin to kolu master via `npins update kolu` once juspay/kolu#1053 merges, before this lands on drishti master.

Workflow completed.

PR #33, opened 2026-05-30
Don't crash at startup when an initial host is unreachable

### LINKED ISSUE #32: Monitor crash-loops at startup when any initial host is unreachable (regression from #31)
## Summary

After updating to `3134fe4` (PR #31, *"Make Reconnect actually re-arm a failed host"*), the monitor **crash-loops at startup and never binds its HTTP port** if *any* host in the initial set is unreachable over SSH. A single offline/unreachable remote takes down monitoring of **all** hosts — including reachable ones (`localhost`, other live hosts) — and the HTTP server (port 7720) never comes up at all.

This is a regression: `9ef2d01` (the previous commit) handled an unreachable host as a soft, per-host retry and kept serving.

## Environment

- drishti: `3134fe4` (broken) vs `9ef2d01` (last good)
- Deployed via the home-manager module (`nix/home/module.nix`) as a **launchd** LaunchAgent on macOS (`aarch64-darwin`)
- `services.drishti.hosts = [ "localhost" "sincereintent" "pureintent" "vanjaram.tail12b27.ts.net" "nix-infra@rasam.tail12b27.ts.net" ]`
- Three of the remotes (`pureintent`, `vanjaram`, `rasam`) are currently unreachable from the launchd context (stale ssh-agent / network) — `ssh`/`nix` to them exits 255.

## Symptom

The launchd agent never stays up; `launchctl print` shows `last exit code = 1` and `state = spawn scheduled` (respawn loop), and nothing ever listens on the configured port:

```
$ curl -s -o /dev/null -w "HTTP %{http_code}\n" http://localhost:7720/
HTTP 000
$ launchctl print gui/$(id -u)/org.nix-community.home.drishti | grep -i 'last exit\|state'
        state = spawn scheduled
        last exit code = 1
```

Logs (`~/Library/Logs/drishti.err.log`) show `localhost` connecting fine, then a `fatal` on the first unreachable host, repeated every respawn:

```
[server] hosts from CLI (5): localhost, sincereintent, pureintent, vanjaram.tail12b27.ts.net, nix-infra@rasam.tail12b27.ts.net
[bridge] pinning HostSession (parent-lifetime ref)…
[host:localhost local] localhost: agent realised at /nix/store/…-drishti-agent
[host:localhost remote] first RPC received — link is live
[bridge] processes: snapshot frame #1 — 946 PIDs (cold-start or reconnect)
[server] fatal: vanjaram.tail12b27.ts.net: `nix-instantiate --eval builtins.currentSystem` exited 255
... (process exits 1, launchd respawns) ...
[server] fatal: pureintent: `nix-instantiate --eval builtins.currentSystem` exited 255
... (loops) ...
```

## Root cause

PR #31 probes each host's nix-system (`builtins.currentSystem`, via `@kolu/surface-nix-host`'s `resolveSystem`, i.e. `nix-instantiate --eval` over SSH) to pick the matching agent `.drv`. The initial seeding does this for **all** hosts with `Promise.all`, which rejects on the first failure:

`packages/app/src/server/hostRegistry.ts:114-124`
```ts
// Parallel initial seeding — per-host arch probes are independent, so
// a single user dialing into five hosts shouldn't pay the round-trip
// serially. `Promise.all` propagates the first failure (Bun's default);
// a failing probe at boot is a misconfiguration we want loud and early.
const seeded = await Promise.all(
  opts.initialHosts.map(
    async (host): Promise<readonly [string, HostHandle]> => [
      host,
      await buildEntry(host),   // -> resolveDrvPath -> resolveSystem(host) over SSH
    ],
  ),
);
```

That rejection propagates up to the top-level handler and exits the process:

`packages/app/src/server/main.ts:248-251`
```ts
main().catch((err) => {
  process.stderr.write(`[server] fatal: ${(err as Error).message}\n`);
  process.exit(1);
});
```

Crucially, `buildHostRegistry` is awaited (`main.ts:95`) **before** the HTTP server is created (`serve(...)` at `main.ts:128`). So a single unreachable host at boot kills the process before the port is ever bound — the admin UI / reconnect surface that PR #31 added is unreachable precisely when you'd want it to re-arm the failed host.

The inline comment frames a boot-time probe failure as *"a misconfiguration we want loud and early"* — but for a long-lived monitor of multiple remotes, a transiently-unreachable host at boot is expected, not a misconfiguration, and shouldn't be fatal for the whole service.

## Expected behavior

An unreachable host at startup should degrade to a per-host **failed/disconnected** state (same as a host that drops later), leaving the HTTP server bound and the other hosts monitored — and let the new Reconnect flow re-arm it once SSH recovers. It should not crash the parent.

## Suggested fix

- Seed with `Promise.allSettled` (or wrap each `buildEntry` so a probe failure yields a "failed" handle rather than throwing), so one bad host can't reject the whole registry; and/or
- Bind the HTTP server **before** seeding hosts, so the admin surface is up even if every host probe fails.

## Reproduction

1. Run the monitor with at least one unreachable host in the set, e.g. `drishti localhost user@unreachable-host`.
2. Observe `[server] fatal: <host>: \`nix-instantiate --eval builtins.currentSystem\` exited 255` and the process exiting 1; the port never binds.

## Workaround

Pin back to `9ef2d01`, or drop unreachable hosts from the set.

srid:
**A single unreachable host at boot no longer takes down the whole monitor.** Previously, one offline remote in the initial host set crash-looped the process and the HTTP port (7720) never bound — so the admin/Reconnect surface meant to recover the host was itself unreachable. Now an unreachable host degrades to a per-host `failed` state (re-armable via the existing **Reconnect** button), every reachable host keeps being monitored, and the port always binds. Regression from #31.

### Root cause

`buildHostRegistry` seeded all hosts by awaiting a per-host arch probe (`resolveSystem` — an ssh `nix-instantiate`) inside `Promise.all`, *before* `serve()` bound the port. `Promise.all` rejects on the first failure, which propagated to `main()`'s top-level catch and `process.exit(1)`:

```
one host's probe throws ─▶ Promise.all rejects ─▶ main().catch ─▶ exit(1)
                                                   (port never bound)
```

A host that drops *later* was already handled gracefully — it had been probed once, so it owned a session whose disconnected→backoff→failed→reconnect machinery caught the drop. A host unreachable *at boot* never got that far: the probe threw before any session existed.

### The fix: move the probe into the one place that already owns SSH reachability

The arch probe now runs *inside* the session's spawn cycle rather than eagerly in the registry, so a probe failure flows through the **same** machinery a later drop does — there's one home for "host unreachable over SSH," not two. This is a companion change in the pinned `@kolu/surface-nix-host` dependency ([juspay/kolu#1054](https://github.com/juspay/kolu/pull/1054)): `HostSession` took a fixed `drvPath: string`; it now takes `resolveDrvPath: () => Promise<string>`, resolved per spawn attempt, with a rejecting resolver handled exactly like a `provisionAgent` failure.

On the drishti side that makes `buildEntry` **synchronous** — it hands kolu the thunk instead of awaiting the probe — so seeding becomes a plain loop that *cannot* reject. `main.ts` needed no change: with no pre-bind host I/O left, the port always binds. The now-dead in-flight `adding` set (it only ever guarded the async-probe window) falls out as a cleanup.

> [!NOTE]
> **Companion change [juspay/kolu#1054](https://github.com/juspay/kolu/pull/1054) has landed on master**, and the kolu pin in `npins/sources.json` now tracks master HEAD (`9b0128d`). CI is green on both x86_64-linux and aarch64-darwin against the merged dependency.

### Try it locally

Reproduce the old crash (now fixed) by pointing it at one reachable and one bogus host:

```sh
nix run github:srid/drishti/fix/startup-unreachable-host-crash -- localhost user@nonexistent.invalid
```

The monitor binds `:7720`, serves `localhost`, and shows the bad host as `failed` with a Reconnect button — instead of exiting 1.

Fixes #32


srid:

Both lenses reviewed `git diff origin/master...HEAD` (the drishti side; the kolu change is a pinned dependency, out of this diff's scope).

| 1 | Hickey | `adding` in-flight set / window — pre-existing, improved by the diff | ⚠️ **No-op** |
| 2 | Hickey | Session pool key drops `drvPath` (upstream kolu) — safe: registry is built once per process | ⚠️ **No-op** |
| 3 | Lowy | `adding`-set comment justified itself with an "async arch-probing" race that no longer exists | Fixed in this PR |

Both confirmed the **boundary move is correct, not a regression**: the "host unreachable over SSH" volatility now lives in exactly one place — the session's existing `disconnected → backoff → failed → reconnect` machinery — instead of being split between a fatal eager probe and the resilient session. No new abstraction, no concept multiplication.

### Hickey rationale

Replacing `await opts.resolveDrvPath(host)` (eager value) with the thunk `() => opts.resolveDrvPath(host)` (lazy callback into `getHostSession`) is the right seam — it folds the probe failure into the session's existing failure path rather than adding a new one. The `Promise.all` → `for` loop change isn't an independent decision: once `buildEntry` can't reject, sequential sync seeding is strictly simpler. Two low-severity observations, both **No-op**: the `adding`/`entries` "host is known" predicate stays centralized in `has()` (no fragmentation), and the pool-key narrowing is safe because the registry is constructed exactly once per process.

### Lowy rationale

The relocated volatility is *"when/how the agent `.drv` is resolved, and what happens when that resolution fails."* Before, it ran outside the session and its failure escaped to `process.exit(1)` with total blast radius; after, it runs at the top of `HostSession.spawn()` where a rejection is treated identically to a `provisionAgent` failure, blast radius one host. The registry→session seam is clean layering — `main.ts` knows the drv map, the registry knows host identity, the session knows neither. One **Fix**: the `adding`-set comment described a removed risk.

### Cross-lens outcome

Lowy recommended *keeping* the `adding` set and fixing its comment; the `/code-police` elegance pass then showed the set is **dead** (with sync `buildEntry`, `adding` is populated and cleared inside one synchronous block no observer can see — it only ever guarded the async-probe window). The deeper call won: the set was removed entirely (`refactor(police): elegance …`), which subsumes Lowy's finding #3. Concurrent same-host `add()` stays safe via the synchronous `entries.has` check before the first await.

srid:
## Evidence

Captured from the real app on this branch: launched with `localhost user@nonexistent.invalid` (one reachable, one unreachable). The monitor **starts and binds its HTTP port** even though one host is unreachable — `localhost` is connected with a live process table (457 PIDs), and the bogus host exhausts its five back-off retries (2 s → 4 s → 8 s → 16 s) and degrades to a per-host **failed** tab. Pre-fix this configuration exited 1 and never bound the port at all.

**localhost — connected, monitoring 457 processes**
localhost tab showing connected state with live htop-style process table

**user@nonexistent.invalid — graceful failed state with Reconnect button**
The tab shows the `● failed` badge, the "Couldn't reach this host" error card with the exact ssh exit-255 message, the full connection log of all five retries, and the **↻ Reconnect** button — the degraded-but-recoverable UX this fix delivers.
bogus host tab showing failed state, connection log, and Reconnect button

srid:

| research | ✓ | 2m 45s | Pinned change sites in kolu `hostSession.ts` + drishti `hostRegistry.ts`; only 2 kolu callers; chose kolu `hostSession.test.ts` as regression-test home |
| branch | ✓ | 8s | On `fix/startup-unreachable-host-crash` off `origin/master`; buggy `Promise.all` seeding confirmed |
| implement | ✓ | 7m 14s | kolu: `resolveDrvPath` thunk + spawn-cycle resolve + pool-key narrowing, new regression test (3/3, workspace typecheck clean); drishti: sync `buildEntry` + seeding loop; re-pinned kolu to PR commit `9f0f70a` + re-hydrated |
| check | ✓ | 34s | `bun run typecheck` clean against new kolu API |
| docs | ✓ | 34s | README: probe now in spawn cycle; boot-unreachable host degrades to per-host failed/Reconnect |
| fmt | ✓ | 25s | `just fmt` clean (nix-only by design) |
| commit | ✓ | 16s | Feature commit pushed |
| hickey+lowy | ✓ | 3m 50s | hickey: 0 findings (2 No-ops); lowy: 1 Fix (stale `adding` comment) → commit `f37ef43` |
| police | ✓ | 6m 50s | rules + fact-check clean; elegance removed the now-dead `adding` set + stale docs → commit `refactor(police)` |
| test | ✓ | 10s | `bun test`: 114 pass / 0 fail; behavior covered by kolu `hostSession.test.ts` |
| create-pr | ✓ | 57s | Draft PR + hickey/lowy ledger comment |
| ci | ✓ | 4m 31s | justci: **all 16 nodes** green on x86_64-linux + aarch64-darwin against HEAD `70afb2b` |
| evidence | ✓ | — | Two real-app screenshots (localhost live + bogus host failed-state) posted |
| **Total** | | **~34m** | |

**Slowest step**: `implement` (7m 14s) — dominated by the cross-repo kolu setup (pnpm install, nix rebuilds of the changed kolu package, re-hydration into `node_modules`).

### Optimization suggestions

- **The cross-repo kolu change is the bulk of `implement` + the merge blocker.** This PR can't land until [juspay/kolu#1054](https://github.com/juspay/kolu/pull/1054) goes green (its `ci::*@aarch64-darwin` checks need a justci Mac runner) and merges to master, after which this needs a one-line `npins` re-pin to master. Landing the kolu PR first would let the final drishti pass run unblocked.
- **For a re-run that only tweaks the drishti side**, use `/do --from post-implement` to skip re-doing the kolu build + hydration (~7m saved).
- **`police` (6m 50s) earned its keep** — its elegance pass caught dead code (`adding` set) that the hickey/lowy pass had kept. No fix-loops; the time was the three passes themselves, not iteration.
- **Local `check` + `test` (~45s combined) caught everything CI did on the linux lane.** When iterating, lean on those before the full justci run (which adds the aarch64-darwin remote lane, ~4m).

> [!NOTE]
> The drishti pin currently tracks the unmerged kolu PR branch (`9f0f70a`); re-pin to master once kolu#1054 lands, before merging this.

Workflow completed.

PR #34, opened 2026-05-31
Timestamp + per-host-tag connection logs to diagnose connect-flapping

srid:
**The parent server's connection logs couldn't be read on a timeline.** There were no timestamps anywhere, and all N per-host bridge loops wrote a single, flat, interleaved `[bridge]` prefix with no host attribution — so a flapping reconnect (every host connects once, then every later reconnect's handshake times out at 30s → `failed`) was impossible to pin from the logs. This enriches the logging so the *next* production capture is diagnostic, then the underlying bridge bug can be fixed with evidence instead of a guess.

The forensic walk-through that motivated this — agent respawns cleanly every time, but the parent stops re-issuing the `system.get` subscription after the first link drop, so the 30s watchdog reaps each idle agent — is in [`docs/plans/talk-drishti-connect-failures.html`](../blob/enrich-connection-logging/docs/plans/talk-drishti-connect-failures.html).

### What the enrichment adds

- **One `makeLogger(tag)` factory** (`server/log.ts`) replacing five hand-rolled `process.stderr.write` writers (`[server]`, `[hosts]`, `[admin]`, `[bridge]`, plus inline writes). Every subsystem — including the host registry (`[registry]`) and the client bundler (`[build]`) — now carries its own tag.
- **Per-host bridge tags** — each host's bridge loop logs `[bridge:<host>]`, so the concurrent loops stop interleaving into one unattributable stream. `buildRouter` builds the tagged logger (the registry passes `host` in, since `HostSession` keeps it private).
- **Timestamps at the sink** — `installStderrTimestamps()` wraps `process.stderr.write` once at boot to ISO-stamp every line. This is the one seam that also reaches `@kolu/surface-nix-host`'s own `[host:X]` writes and the forwarded remote-agent stderr, which a per-writer prefix can't. On Linux/systemd, journald already stamps.
- **Reconnect-handoff instrumentation** (`pumpSystemCell`) — a per-spawn client sequence id, an `issuing system.get subscription (client #N)` line, and elapsed-to-first-RPC on the first yield. A subscription that never roundtrips on a new client becomes visible and distinct from "the bridge never got a new client".
- **Agent-side liveness** — pid in `first RPC received`, and a `waiting for first RPC (Ns)…` heartbeat while blocked on stdin, so a reaped idle agent proves the stall is the parent→agent handshake, not the remote.

### What a stuck reconnect now looks like

```
19:01:42.118 [bridge:sincereintent] agent client ready (client #7); starting pumps
19:01:42.119 [bridge:sincereintent] system: issuing system.get subscription (client #7)
19:01:44.301 [host:sincereintent remote] serving surface over stdio
19:01:54.301 [host:sincereintent remote] waiting for first RPC (10s)…
19:02:12.119 [host:sincereintent local] connect handshake timed out after 30000ms
```

The stalled bridge, its client id, and the agent's idle wait are all attributable and on a timeline — which the flat untimed log could not show.

> _No behavioral change to the connection lifecycle itself — this is diagnostics only. The bridge bug it surfaces is a follow-up, to be fixed against the richer log._

### Try it locally

```sh
nix run github:srid/drishti/enrich-connection-logging -- localhost
```

### DIFF
--- a/.agency/do.md
+++ b/.agency/do.md
@@ -10,13 +10,10 @@ just fmt
 bun test
 
 ## CI command
-nix run github:juspay/justci -- run --host x86_64-linux=localhost
+nix run github:juspay/justci -- run
 
-<!-- `--host x86_64-linux=localhost`: the configured `srid-justci` SSH alias
-routes through a Tailscale proxy that returns "not owner of srid-justci",
-so the linux lane runs against this machine until the proxy is restored.
-Drop the override once `~/.config/justci/hosts.json` points linux somewhere
-reachable. -->
+<!-- Host routing comes from `~/.config/justci/hosts.json`
+(x86_64-linux → vanjaram, aarch64-darwin → rasam); no `--host` override. -->
 
 
 ## Documentation
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ The monitor binds `0.0.0.0` on `port`. `hosts` are passed as positional argument
 
 See [`nix/home/example/`](nix/home/example/) for a full configuration — a NixOS VM test exercises the systemd path on Linux, and a standalone home-manager activation build exercises the launchd path on Darwin.
 
-On macOS, the LaunchAgent writes stdout to `~/Library/Logs/drishti.out.log` and stderr to `~/Library/Logs/drishti.err.log`, so crashes and startup failures leave logs alongside other user logs.
+On macOS, the LaunchAgent writes stdout to `~/Library/Logs/drishti.out.log` and stderr to `~/Library/Logs/drishti.err.log`, so crashes and startup failures leave logs alongside other user logs. Every stderr line is timestamped (ISO-8601) and tagged by subsystem — `[server]`, `[hosts]`, `[admin]`, and one `[bridge:<host>]` per monitored host — so a connection problem can be traced to a specific host on a timeline (on Linux/systemd, `journalctl --user -u drishti` adds its own timestamps).
 
 ## Project layout
 
--- a/docs/plans/talk-drishti-connect-failures.html
+++ b/docs/plans/talk-drishti-connect-failures.html
@@ -1,0 +1,162 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>drishti — why 3 hosts fail to connect</title>
+<style>
+  :root{
+    --bg:#0f1419; --panel:#161b22; --panel2:#1c232c; --ink:#d7dee6; --dim:#8b97a5;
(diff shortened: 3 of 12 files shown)


srid:

| 1 | Hickey + Lowy | `installStderrTimestamps` not idempotent — a second call wraps the already-wrapped `write`, double-stamping every line | Fixed in this PR |
| 2 | Lowy | Agent log lines carry no bracket tag like parent subsystems do | ⚠️ **No-op** |
| 3 | Lowy | Host-registry lifecycle events logged under `[server]`, not their own tag | Fixed in this PR |
| 4 | Lowy | `build.ts` used untagged `console.error` | Fixed in this PR |
| 5 | Lowy (cross-validate) | Relocate `installStderrTimestamps` into `main.ts` instead of guarding it | ⚠️ **No-op** |
| 6 | Lowy | `installStderrTimestamps` is on the correct (sink) side of the volatility line | ⚠️ **No-op** |

### Hickey rationale

The diff does the intended dedup correctly — `log.ts` introduces no pattern that already existed, and collapsing the five hand-rolled writers into one `makeLogger(tag)` factory repairs pre-existing fragmentation rather than reintroducing it. The per-host bridge logger is threaded as an explicit parameter (not closure-over-global), and `makeLogger`/`installStderrTimestamps` are cohesive (both concern "how this process writes to stderr").

The one finding to land: **`installStderrTimestamps` was not structurally idempotent** — a second call captured the first wrapper as `original`, producing a silent double-stamping chain. Fixed with a `Symbol`-keyed marker on the installed wrapper (survives the test's restore-and-reinstall pattern), making idempotency structural rather than a "call it once" discipline.

### Lowy rationale

The §1 duplication check passes: no prior `Logger`/`makeLogger` seam existed, so this is the correct *first* encapsulation of the "logging format/routing" axis, with five real consumers on day one. `installStderrTimestamps` at the parent sink is the **only** layer that can reach kolu's direct `[host:X]` writes and the forwarded agent stderr — correct side of the volatility line (finding 6, No-op).

Three attribution-asymmetry findings: the registry tag (3) and `build.ts` tag (4) were fixed — every subsystem now self-attributes without the caller threading a logger in. The agent `[agent]` tag (2) was **rejected as No-op**: cross-validation (verified against `@kolu/surface-nix-host` `hostSession.ts`) confirmed kolu already prefixes every forwarded agent line with `[host:X remote]`, so a local `[agent]` tag would *double-tag* — the premise that agent lines arrive bare is false. The useful disambiguator (`pid=`) is already added.

### Cross-validation

Ran both lenses against each other's findings. Hickey confirmed Lowy's `[agent]` recommendation (2) would complect agent authorship with kolu's transport attribution → kept as No-op. Lowy proposed relocating `installStderrTimestamps` into `main.ts` as an alternative to the idempotency guard (5); the two are mutually exclusive resolutions of the same concern — the `Symbol` guard was chosen since it keeps the stamping logic unit-testable and the two `log.ts` functions are cohesive (Hickey's cross-val). No conflicts survived.

PR #41, opened 2026-05-31
Scope the agent to its own build so rebuilds keep remotes warm (#38)

### LINKED ISSUE #38: Every reconnect after a drishti rebuild re-copies + rebuilds the full agent closure on each remote
## Summary

The agent `.drv` is baked per drishti build (`DRISHTI_AGENT_DRVS_JSON` from `flake.nix`). It's stable *within* one build, but **every time drishti is rebuilt/redeployed the drv hash changes**, so the first reconnect to each remote afterward is a full cross-arch **rebuild + closure copy** on the target — not the fast no-op a cached agent should be. That long, fragile provisioning window is what transient network problems get caught in.

## Evidence (from production logs)

The agent drv hash walks across the day, changing with each redeploy:

```
…zpydy5wq…-drishti-agent.drv   → …cjhi94xw…  → …3qslrg3h…  → …n8pqds2q…
→ …75za9myp… → …63a9s9il… → …8xhfzhld…
```

Each *new* hash means vanjaram has never seen it → `nix copy --derivation` ships the closure and `nix-store --realise` builds it remotely. Observed: a single fresh-drv provision spent **~10 minutes** just copying (`copying derivation …8xhfz…` 15:49:18 → `copying 5 paths…` 15:59:49) before realising — and that's the window that wedged for hours when the host's ssh degraded mid-realise (see kolu#1071, which stops the *hang*; it does not shorten the *window*).

Within a stable build the drv is reused (`cjhi94` served reconnects across ~11 hours), confirming the cost is tied to **rebuild churn**, not per-connect non-determinism.

## Why it bites

- We rebuild drishti often (active development + launchd `KeepAlive` restarts on crash). Each redeploy resets every remote to a cold cache.
- The provision runs **lazily on the first reconnect** of each host, so the cost lands at the worst time (a host trying to recover) rather than up front.
- A slow/degraded link during that window stalls the connect. kolu#1071 makes a dead peer fail-fast (~30s) and retry, but a healthy-but-slow 10-minute copy is still 10 minutes of "Copying agent to remote…".

## Possible directions (not committed)

1. **Pre-warm on startup / on rebuild:** copy+realise the current agent drv to all remotes proactively (with its own progress surface), so reconnects find a warm cache instead of paying the cost lazily.
2. **Don't churn the drv when it needn't change:** scope the agent derivation so unrelated drishti changes (client/server-only edits) don't alter the *agent* closure's hash — minimizing how often remotes go cold.
3. **Survive the redeploy:** if the new agent drv is unchanged for a host, skip re-provisioning entirely (the GC-root pin already keeps the old output alive — confirm the fast path is actually hit on reconnect).
4. **UI honesty:** distinguish "rebuilding agent on remote (cold)" from a quick reconnect so a 10-minute first-connect doesn't look like a hang.

## Related

- kolu#1071 — keepalive so a degraded host can't wedge the realise *forever* (fixes the hang; this issue is about the *cost/frequency* of the realise itself).
- The reconnect saga: `docs/blog/the-reconnect-that-wedged-then-spun.md`.

srid:
**The remote `drishti-agent` now builds from its own minimal Nix derivation, so a drishti rebuild no longer re-copies the agent closure to every host.** The agent `.drv` used to be fused to `drishtiBuilt` — the whole app tree (SolidJS client + Bun parent server + `dist` bundle + full `node_modules`) — so *any* edit rotated its hash, and the first reconnect to each remote afterward paid a full cross-arch `nix copy --derivation` + `nix-store --realise` (~10 min, observed). That long, fragile provisioning window is what transient network problems got caught in (see the postmortem in `docs/blog/`). Now the agent's inputs are scoped to just the agent, so client/server-only rebuilds leave its `.drv` byte-identical and every remote keeps its warm cache.

Closes #38. Design rationale: `docs/plans/plan-issue-38-agent-drv-churn.html`.

### Two churn edges, both cut

The workspace split alone was *not* enough: bun2nix's dep cache is a `symlinkJoin` over **every** member's source FOD, so it had to be filtered too. Both edges carried `packages/app` source into the agent `.drv`:

| Edge | Before | After |
|---|---|---|
| `src` fileset | whole `packages/app` (client + server + dist) | `packages/agent` + `packages/common` only |
| bun2nix cache | full `bun.nix` incl. `"drishti-app" = copyPathToStore ./packages/app` | same `bun.nix` with the `drishti-app` FOD removed via `removeAttrs` |

### The split

`packages/app` was the lone workspace member; it's now three, so the agent's `copyPathToStore` FOD is naturally agent-only:

```
packages/
├─ common/   drishti-common — per-host wire contract (surface.ts), shared by agent + monitor
├─ agent/    drishti-agent  — main.ts + proc.ts, built by the new scoped derivation
└─ app/      drishti-app    — client + parent server (behavior unchanged)
```

The monitor + client still build from the full `drishtiBuilt`; only the `drishti-agent` wrapper repoints to the new `drishtiAgentBuilt`. Cross-arch pure eval of the per-system drv map is preserved (no IFD), and there's still a single `bun.lock`.

### Verifying it sticks

A new `just drv-stability` CI node replays a client edit in a throwaway git worktree and asserts the agent `drvPath` is byte-identical — the regression guard for this whole change. It's green locally, and the built agent still serves over stdio. Dropping the `dist`/client build also shrinks the closure (no babel/tailwind/hono/client libs), and right-sizing the deps during review dropped `solid-js` from the agent closure too.

> The agent `.drv` *will* still change — correctly — on a genuine agent change: editing `agent/` or `common/` source, or a `bun`/`@kolu/surface`/lockfile bump. Those are now the only cases that warrant a re-copy.

### Try it locally

```sh
nix run github:srid/drishti/scope-agent-derivation
```


srid:

Both lenses ran as parallel sub-agents on the diff, then a cross-validation pass had each lens audit the other's recommendations. The cross-validation was load-bearing here — it caught that the headline Lowy fix was unsafe as written.

| 1 | Lowy | `drishti-common` declared `@kolu/surface`'s whole dep set (incl. `solid-js`) — bloat the agent never imports | Fixed in this PR (`93bbf38`) |
| 2 | Hickey | `removeAttrs` exclude-list defaults to unsafe for a future `packages/*` member | Fixed in this PR (`e2f7962`) |
| 3 | Hickey | Three byte-identical per-package `tsconfig.json` with nothing marking the duplication intentional | Fixed in this PR (`0d3c386`) |
| 4 | Hickey ×-val | Lowy #1 applied naively breaks the agent at runtime — `@orpc/client` is load-bearing (peer-server → stdio-codec) | Fixed in this PR (folded into `93bbf38`) |
| 5 | Lowy ×-val | Hickey #2 as an *allowlist* / Hickey #3 as *project references* would re-braid what the split separated | Fixed in this PR (constrained both to comment-only fixes) |
| 6 | Hickey | `typecheck` serial order (`common && agent && app`) is convention, not toolchain-enforced | ⚠️ **No-op** |
| 7 | Lowy | `drishti-agent`/`drishti` Nix derivations share ~40 lines | ⚠️ **No-op** |
| 8 | Lowy | `admin-surface.ts` (a 2nd wire contract) sits outside the `drishti-common` receptacle | ⚠️ **No-op** |
| 9 | Lowy | `metrics.ts` / `history.ts` placement | ⚠️ **No-op** |

### Hickey rationale

The new `nix/packages/drishti-agent/default.nix` is **legitimate divergence, not concept multiplication** versus the canonical monitor build (`nix/packages/drishti/default.nix`): they differ in `buildPhase` (agent has none), kolu inputs, `bunDeps`, `src`, and install guards — a shared factory would have to braid "bundle or don't" × "which kolu packages" × "which fileset", complecting the two more than two flat files do (**No-op #7**). Three real findings: the `tsconfig` triplication had no marker of intent (**#3** — comment, not project-references, which would add machinery), and the `removeAttrs` exclude-list silently re-admits any future member into the agent cache while `drv-stability` only probes a client edit (**#2** — invariant comment at the call site). Cross-validating Lowy, the `solid-js` removal is safe but `@orpc/client` is **not** — `@kolu/surface/peer-server → ./links/stdio-codec` imports it, and the agent build excludes `packages/app`, so `drishti-common`/`drishti-agent` are the only declarers (**#4**).

### Lowy rationale

The three new boundaries each encapsulate one volatility axis cleanly: `drishti-common` = the wire contract (poll-cadence schema), `drishti-agent` = OS metric collection + deployment-topology (ships separately, must stay byte-stable), and `drishti-agent/default.nix` = *what source churns the agent hash* (anchored by `drv-stability`). The one misplacement was `drishti-common` declaring `solid-js`/`@orpc/client`/`@solid-primitives` — deps `surface.ts` never imports, tracking "@kolu/surface's package.json" rather than "what this module needs" (**#1**). Fixed by tracing the agent's actual reachable entrypoints and splitting deps per consumer (contract needs → common; server/peer-server needs → agent), dropping `solid-js` from the shipped closure. `admin-surface.ts` is a second wire contract but its blast radius is contained within `packages/app` today, so relocating it now would be speculative (**No-op #8**); `metrics`/`history` are correctly monitor-internal (**No-op #9**). Cross-validating Hickey, neither the allowlist nor project-references form was adopted — both would have leaked package-graph volatility across files, so the comment-only resolutions were taken instead (**#5**).

srid:
## Evidence

This is a **build-only** change — it alters how the `drishti-agent` derivation is *built*, not anything the agent observes or the UI renders. The live process table, charts, and tabs are byte-for-byte unchanged, so there's no UI surface to screenshot. The meaningful "did it work" evidence is the build-level invariant this PR exists to establish, verified in strict-mode CI on both platforms against `HEAD` (`0d3c386`):

**The agent `.drv` no longer changes when client/server code does** (`just drv-stability` — snapshots `drishti-agent.drvPath`, replays a client edit in a throwaway worktree, asserts byte-identical):

| Platform | Builder | `drv-stability` | agent build (`ci::nix`) |
|---|---|---|---|
| `x86_64-linux` | fresh box | ✅ **PASS** | ✅ builds + serves over stdio |
| `aarch64-darwin` | rasam | ✅ **PASS** | ✅ builds + serves over stdio |

> `drv-stability: PASS — agent .drv unchanged across a client edit (…-drishti-agent.drv)` — emitted identically on both lanes (the guard pins the `x86_64-linux` drvPath, so pure cross-arch eval yields the same hash regardless of evaluator).

Full run: **all 18 nodes succeeded** (setup + `_prefetch-crates`, `install`, `nix`, `typecheck`, `fmt-check`, `bun-nix-fresh`, `home-manager`, `drv-stability` × 2 platforms). The `home-manager` linux lane boots the nixosTest VM and the agent still builds with the trimmed, scoped closure (no `solid-js`, no client bundle).

PR #45, opened 2026-05-31
Link the parent PID in the process detail panel

srid:
**The PARENT row in a process's detail panel is now a clickable link** that jumps the panel straight to the parent process — selecting its row and swapping the open detail panel to it in one click. Walking a process tree (a `claude` parent → its `chrome-headless-shell` children, say) no longer means scanning the table by eye for the ppid.

The `ppid` was already flowing schema → agent → wire → renderer and rendering as plain text; this only makes it actionable. Clicking re-points the same `selectedPid` signal that already drives row highlighting and the panel, so there's no new selection state — navigating to the parent reuses exactly the path a row click takes.

The link only appears when the parent is **still in the live set**. For pid 1 / orphans (`ppid` 0), or a parent that has since exited, the ppid stays plain text — selecting an absent pid would just close the panel, so the affordance is hidden rather than dangling.

> The two structural-review commits on this branch are worth a glance: the linking action is wired as an *unconditional* `setSelectedPid`, deliberately **not** the `toggle` the rows use (which closes on re-select) — navigating to a parent must always re-point, never close. The second commit collapses the two props into one nullable `onSelectParent: (() => void) | null`, making the "no parent, but still callable" state unrepresentable.

### Try it locally

```sh
nix run github:srid/drishti/link-parent-pid
```

Open a host, click a process row, then click its **parent** value.


srid:

Both lenses ran on the primary commit, then a second parallel pass cross-validated each lens against the *other's* recommendations. The two first-pass fixes actually conflicted — Hickey wanted to keep `ProcessDetail` a pure prop-driven renderer and just unify the verb; Lowy wanted to fold selection into `SelectionContext`. Cross-validation resolved the conflict.

| 1 | Hickey | `parentPresent` liveness invariant holds via Solid fine-grained reactivity — no stale-link bug | ⚠️ **No-op** |
| 2 | Hickey | No new concepts, signals, or structural entanglement; `ProcessDetail` keeps its pure-renderer shape | ⚠️ **No-op** |
| 3 | Hickey + Lowy | Name the *unconditional-set vs `toggle`* distinction at the selection boundary | Fixed in this PR |
| 4 | Hickey (cross) | Reject the `onSelectPid={toggleSelected}` swap — it would regress to close-on-navigate | ⚠️ **No-op** |
| 5 | Lowy | Migrate `onSelectPid` into `SelectionContext` to dedupe the selection receptacle | ⚠️ **No-op** |
| 6 | Lowy (cross) | `parentPresent` is on the *liveness* axis — keep it as data, never fold into the selection context | ⚠️ **No-op** |

### Hickey rationale

The diff adds no new files, signals, or lifecycle scopes — a pure extension of the existing `ProcessDetail`. The liveness invariant (`link shown ⇔ parent in store`) is held mechanically by reactivity, not by convention: the guard expression lives in a reactive `<Show>` scope and re-tracks the `processes` store, so a parent that exits degrades the link to plain text on the next tick. A forked run mis-flagged this as a "stale snapshot" bug; verified against `babel-preset-solid` + the reactive consumer, it's a live getter — **No-op**.

The one real wrinkle: the link selects via the bare `setSelectedPid` while rows select via `toggle`. The first instinct was to unify both onto `toggle` — but cross-validation showed that's a **regression**: once selection has moved, a parent link can point back at the currently-open pid, where `toggle` would *close* the panel instead of staying on it. So the unconditional set is the correct verb; the fix is to *name* the distinction in a comment, not erase it (commit `e8f8ced`).

### Lowy rationale

Selection-mutation is one volatility axis, and `SelectionContext` is its declared receptacle (its own comment says it exists to avoid widening props). The first-pass call was to fold `onSelectPid` into the context. Cross-validation walked this back: `ProcessDetail` is a **sibling** of the table, not buried in its subtree, so there's no prop-drilling depth to remedy; migrating would *grow* the context interface (a new `select` member reachable by every `ProcessRow` that doesn't need it) rather than shrink concept count. And `parentPresent` is on a **different** axis (process-set liveness, derived from the `processes` store) — folding it into the selection context would complect the two. Both halves therefore stay as data passed to a directly-rendered sibling. The valid kernel of the finding — the unnamed verb asymmetry — is resolved by the same doc commit as Hickey #3.

_The elegance pass later collapsed the two props into a single nullable `onSelectParent` (`7595693`), making the "no parent yet still callable" state unrepresentable — which tightens the boundary further without touching the axis decision above._

srid:
## Evidence

Parent PID is now a link in the detail panel — clicking it jumps to the parent process.

**Before — child process selected (PID 781506), parent PID 781402 rendered as an emerald clickable link:**
parent link

**After clicking the parent PID — panel swapped to PID 781402, its row highlighted, and the panel's own PARENT field now shows 781364 as the next link up the chain:**
after click

srid:

| sync | ✓ | 1s | forge=github; noGit=false |
| research | ✓ | 1m 52s | ppid already wired; PARENT row at App.tsx:1281; selection guards on store presence |
| branch | ✓ | 4s | feature branch link-parent-pid off origin/master |
| implement | ✓ | 33s | ppid renders as a button when parent present, plain text otherwise |
| check | ✓ | 25s | typecheck (tsc --noEmit) clean across common/agent/app |
| docs | ✓ | 17s | README.md:40 updated — parent PID is a clickable link |
| fmt | ✓ | 15s | nixpkgs-fmt clean (0/13); no JS formatter wired in repo |
| commit | ✓ | 10s | 6d68f93 pushed |
| hickey+lowy | ✓ | 7m 47s | ran + cross-validated; Hickey toggle-swap rejected (regression), Lowy context-migration declined; doc-only fix e8f8ced |
| police | ✓ | 3m 51s | 3 passes clean; elegance fix (collapse to onSelectParent) 7595693 |
| test | ✓ | 28s | bun test 119 pass / 0 fail |
| create-pr | ✓ | 1m 11s | draft PR + hickey/lowy ledger comment |
| ci | ✓ | **13m 6s** | all 18 justci nodes green on HEAD (both lanes incl. NixOS VM test); status=success |
| evidence | ✓ | 3m 22s | 2 screenshots posted (parent link + after-click swap) |
| **Total** | | **33m 39s** | |

**Slowest step**: `ci` (13m 6s)

### Optimization suggestions

- **`ci` dominated at 13m (39% of wall-clock)** — and most of that was a detour: the first strict run wedged because the configured `x86_64-linux` host (`vanjaram`) was unreachable, forcing a kill + re-route to `drishti-ci` + full re-run. **Sanity-check `~/.config/justci/hosts.json` reachability before invoking `/do`** (or pin the lane to a known-live box) to avoid the dead-host stall; re-runs can use `--from ci-only`.
- The `home-manager` NixOS **VM test** is the heaviest single recipe per lane (boots a QEMU VM + curls the service). It's the right coverage for a UI-rendering change, but if iterating, `run typecheck@x86_64-linux fmt-check@x86_64-linux` covers the TS-relevant lanes far faster.
- **`hickey+lowy` (7m 47s)** ran a second cross-validation pass because the two lenses' first-pass fixes genuinely conflicted (keep-prop vs. move-to-context). That spend paid off — it caught that Hickey's `toggle` swap would have been a close-on-navigate regression — but for an obviously-confined diff, `--minimal` skips this pass entirely.

Workflow completed.

PR #46, opened 2026-06-01
Serve the agent surface before enumerating processes

srid:
**The `localhost` tab could fail to connect — `connect handshake timed out after 30000ms (transport up, no first RPC)` — even though localhost needs no network at all.** drishti treats every host uniformly: even localhost realises the agent `.drv` and spawns it as a local stdio child, then waits ≤30s for the first RPC. But the agent enumerated *every* process before it started serving — on macOS that's `ps` + `lsof -nP -d cwd` over the whole table — so the parent's first `system.get` sat unread in the child's stdin until that scan finished. On a busy, high-process-count box (a 956-process Mac at load 9.58) the scan overran the 30s connect watchdog, and the host went terminal-`failed`. The irony: localhost is the *worst* case in a fleet — the agent runs on the saturated box itself, while idle Linux remotes (pure `/proc` reads, no `lsof` fork) handshake sub-second.

The fix is **serve first, enumerate lazily** — the handshake needs only the cheap `system` cell, not the `lsof`-driven process table:

```
before:  spawn ─▶ readSystem ─▶ ps + lsof×N ─▶ netstat ─▶ serve ─▶ answer   ✗ >30s on a loaded box
after:   spawn ─▶ readSystem ─▶ serve ─▶ answer   ✓ sub-second
                                  └─ poll tick fills processes / cpu / net, off the critical path
```

### What changed

- Seed only the cheap `system` cell (vm_stat + statfs on darwin; a couple of `/proc` reads on linux) plus the synchronous `os.cpus()` core baseline before `serveOverStdio()`.
- Start the process and network maps **empty**; the poll loop's first tick — kicked immediately, fire-and-forget — fills them, so the `ps`/`lsof`/`netstat` forks never sit on the handshake's critical path.
- Extract `serveAgent(reader, serve)` with an injectable transport and guard the entrypoint with `import.meta.main`, so a regression test can prove the first RPC is answered *while* `readProcesses()` is still blocked.

_First-tick cost: a browser that subscribes in the sub-second window before tick #1 sees an empty process list and 0 B/s rates for one poll, corrected on the next delta — benign under the existing snapshot-then-delta contract._

### Review refinements (hickey / lowy / police)

| Lens | Finding | Fix |
| --- | --- | --- |
| Hickey | `readNetwork()` awaited serially after the `Promise.all` in `tick` | folded into the tick-top `Promise.all` — netstat rides alongside system/processes |
| Lowy + Hickey | injectable type exported as `ServeOverStdio` — names the *implementation*, not the role, with no external consumer | module-private `Serve`; the test passes an inline, contextually-typed fake |
| Police | needless `[...keys()]` spread on the process eviction loop; a vacuous test assertion | both removed |

Full root-cause + design write-up: `docs/plans/talk-localhost-handshake-timeout.html`.

### Try it locally

```sh
nix run github:srid/drishti/fix/agent-serve-before-enumerate
```

### DIFF
--- a/docs/plans/talk-localhost-handshake-timeout.html
+++ b/docs/plans/talk-localhost-handshake-timeout.html
@@ -1,0 +1,214 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>“localhost” timed out — the host that needs no connection</title>
+<style>
+  :root {
+    --bg: #0b1020; --panel: #121a2e; --panel2: #0e1526; --line: #1f2a44;
+    --ink: #d7deec; --dim: #8a96b0; --faint: #5d6886;
+    --amber: #f59e0b; --emerald: #10b981; --red: #ef4444; --blue: #6366f1;
+    --mono: ui-monospace, "SF Mono", "JetBrains Mono", Menlo, monospace;
+  }
+  * { box-sizing: border-box; }
+  body {
+    margin: 0; padding: 2.2rem clamp(1rem, 4vw, 3.5rem); background: var(--bg);
+    color: var(--ink); font: 15px/1.6 system-ui, -apple-system, sans-serif;
+  }
+  main { max-width: 920px; margin: 0 auto; }
+  h1 { font-size: 1.5rem; margin: 0 0 .2rem; }
+  h2 { font-size: 1.05rem; margin: 2.2rem 0 .6rem; color: #fff;
+       border-bottom: 1px solid var(--line); padding-bottom: .35rem; }
+  h3 { font-size: .92rem; margin: 1.3rem 0 .4rem; color: var(--ink); }
+  p { margin: .55rem 0; }
+  code { font-family: var(--mono); font-size: .85em; color: #e6c07b;
+         background: #1a2238; padding: .08em .35em; border-radius: 4px; }
+  .sub { color: var(--dim); margin-top: .1rem; }
+  .cite { font-family: var(--mono); font-size: .8em; color: var(--faint); }
+  .pill { display: inline-block; font-size: .72rem; font-family: var(--mono);
+          padding: .1em .5em; border-radius: 999px; border: 1px solid var(--line);
+          color: var(--dim); }
+  .key { background: #161f38; border: 1px solid var(--line); border-radius: 10px;
+         padding: 1rem 1.2rem; margin: 1rem 0; }
+  .key strong { color: #fff; }
+  .log { background: #07060b; border: 1px solid var(--line); border-radius: 8px;
+         padding: .9rem 1rem; font-family: var(--mono); font-size: .76rem;
+         line-height: 1.55; overflow-x: auto; white-space: pre; color: #aeb6c8; }
+  .log .l { color: var(--dim); }
+  .log .r { color: #7f8aa6; }
+  .log .ok { color: var(--emerald); }
+  .log .warn { color: var(--amber); }
+  .log .stuck { color: var(--red); font-weight: 600; }
+  ul { margin: .5rem 0; padding-left: 1.2rem; }
+  li { margin: .3rem 0; }
+  table { width: 100%; border-collapse: collapse; margin: .8rem 0; font-size: .88rem; }
+  th, td { text-align: left; padding: .5rem .6rem; border-bottom: 1px solid var(--line);
+           vertical-align: top; }
+  th { color: var(--dim); font-weight: 600; font-size: .78rem; text-transform: uppercase;
+       letter-spacing: .03em; }
+  td code { white-space: nowrap; }
+  .flow { font-family: var(--mono); font-size: .82rem; line-height: 1.9; margin: .6rem 0;
+          color: var(--dim); }
+  .flow b { color: var(--amber); }
+  .flow .g { color: var(--emerald); }
+  .flow .x { color: var(--red); }
+  .mockwrap { display: grid; grid-template-columns: 1fr 1fr; gap: 1rem; margin: 1rem 0; }
+  @media (max-width: 720px) { .mockwrap { grid-template-columns: 1fr; } }
+  .mock { background: #0a0f1e; border: 1px solid var(--line); border-radius: 10px;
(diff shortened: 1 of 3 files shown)


srid:

Ran as parallel sub-agents on the diff, then a cross-validation pass (each lens auditing the other's recommendations). All findings actioned — no defers.

| --- | --- | --- | --- |
| 1 | Hickey | `readNetwork()` awaited serially after the `Promise.all` in `tick` | Fixed in this PR (`6d23c11`) |
| 2 | Hickey | process-eviction loop spreads `[...keys()]` while siblings don't — claimed removal-skip risk | ⚠️ **No-op** (verified spec-safe; later removed for consistency by police) |
| 3 | Lowy + Hickey | injectable type exported as `ServeOverStdio` — names the *implementation*, not the role; no external consumer | Fixed in this PR (`15ae6c0`) |
| 4 | Lowy | `serveAgent` is shaped as a testability seam, not a volatility boundary | ⚠️ **No-op** (the seam earns its keep) |
| 5 | Lowy | `import.meta.main` entrypoint guard | ⚠️ **No-op** (correct hygiene, mirrors `build.ts`) |
| — | both | the serve-before-enumerate reorder itself | ⚠️ **No-op** (it is the right cut) |

_Cross-validation converged: Lowy-cross affirmed Hickey's `readNetwork` fold is pure sequence-volatility with no coupling; Hickey-cross agreed `serveAgent` must stay exported (Bun can't import a non-exported symbol) and refined the type finding to the lowest-artifact resolution that was applied — drop the export, make it a module-private `Serve` named for the role._

### Hickey rationale

The reorder decomposes a startup pipeline along its real dependency seam: the handshake needs only the cheap `system` cell and the synchronous `os.cpus()` baseline, so those seed before `serve()`; the expensive `ps`+`lsof`+`netstat` reads start as empty maps and are filled by the poll loop's first tick. No concept is multiplied; the new `Serve` type is a minimal test seam, and `import.meta.main` is the idiomatic entrypoint guard.

**Finding A (fixed):** `readNetwork()` is an independent forked probe (darwin `netstat`) with no dependency on the process-diff, yet it was awaited serially after `Promise.all([readSystem, readProcesses])` — making each tick `max(system, processes) + network`. Folded into the `Promise.all`.

**Finding B (no-op):** the eviction loops delete only the *current* key during Map iteration, which is spec-safe (an empirical check confirmed every expected removal still fires). Not a correctness bug — so the original "add the spread everywhere" framing was dropped; police instead removed the lone spread for consistency.

### Lowy rationale

The volatility being respected is **latency-on-the-critical-path vs. off-it**, and the reorder honors it: `ProcReader` (in `proc.ts`) encapsulates *how* each metric is read; `tick` owns *when/in what order* — the parallelism decision is pure sequence volatility and lives in one place.

**Export/naming (fixed):** `serveAgent`/`ServeOverStdio` had exactly one consumer (the co-located test) and no barrel re-export, so the `export` advertised a package API nothing external consumes — and `ServeOverStdio` named the one production implementation rather than the seam's role. Resolved by making the type module-private and renaming it `Serve`; `serveAgent` stays exported because the test imports it.

**No-ops:** the `serveAgent` extraction (a justified observation seam, not over-decomposition), the `import.meta.main` guard, and the reorder itself.

srid:
## Evidence

Captured on a live `localhost` run of the reordered serve-first agent (PR #46): the parent connected successfully, the agent enumerated the process table, and the browser received the full snapshot — confirming that serving the RPC surface *before* running `ps`/`lsof` does not break normal operation end-to-end.

default load

465 PIDs visible on `localhost` (x86_64-linux); header shows load avg 36.19 / 16.13 / 6.93, MEM 14.7 / 29.2 GB, uptime 3d 11h.

_Note: this is a normal-operation / no-regression capture. The original timeout reproduces only on a heavily-loaded macOS box (where `lsof` over a large process table overran the 30s handshake); on a Linux host localhost always connected sub-second, so a before/after timeout shot isn't reproducible here — the full `justci` pipeline (incl. the home-manager VM boot/curl test) is the end-to-end gate._

srid:

| sync | ✓ | 0s | forge=github, default=master |
| research | ✓ | 0s | root cause + design in `docs/plans/talk-localhost-handshake-timeout.html` |
| branch | ✓ | 0s | `fix/agent-serve-before-enumerate` off origin/master |
| implement | ✓ | — | serve-first reorder + `serveAgent` extraction + regression test |
| check | ✓ | 40s | `tsc --noEmit` clean (common+agent+app) |
| docs | ✓ | 18s | README connect/watchdog prose still accurate — no edit needed |
| fmt | ✓ | 44s | `nixpkgs-fmt` clean (project has no Biome gate) |
| commit | ✓ | 18s | plan doc + fix pushed |
| **hickey+lowy** | ✓ | **11m 10s** | parallel review + cross-validation; 2 fixes, rest No-op |
| test | ✓ | 53s | `bun test` 27 pass / 0 fail; new behavior covered |
| police | ✓ | — | rules/fact-check/elegance all clear (2 elegance fixes) |
| create-pr | ✓ | 1m 39s | draft PR + Hickey/Lowy comment |
| ci | ✓ | 2m 55s | `justci` 18/18 green (x86_64-linux + aarch64-darwin) |
| evidence | ✓ | 4m 24s | default-load screenshot (465 PIDs) |
| **Total** | | **~36m** | |

**Slowest step**: `hickey+lowy` (11m 10s)

### Optimization suggestions

- **Cross-validation doubled the review latency** (the 11m long pole). For a diff this small — one substantive file — a single hickey+lowy pass would likely have surfaced the same two findings; reserve the cross-validation round for multi-file or boundary-heavy diffs.
- **The evidence step rebuilt the client bundle + drove a browser for a backend-only change** (4m 24s). The `home-manager` CI node already boots a VM and curls the service end-to-end, so for agent/server-only diffs the browser capture is low marginal value — `--minimal` would skip it (plus docs/police/structural review) on trivially-scoped changes.
- **`check`/`fmt` each paid ~40s of `nix develop` cold entry.** Pre-warming the devshell (`nix develop -c true`) before invoking `/do` collapses that to near-zero on the later steps.
- For a follow-up tweak after review, **`--from ci-only`** re-runs just the (warm, ~3m) pipeline against the new HEAD.

Issue #43, opened 2026-05-31
Preferred private channel for a security report?


endscene665:
Hi, I found a security issue in the current drishti tree involving the default network exposure and the parent monitor RPC surface. I have not tested any deployed service. I would like to send a compact source-level report and patch outline privately. What channel should I use?

srid:
@endscene665 Sorry just saw this. You can email me: srid@srid.ca

srid:
I'll close this because I believe you were referring to https://github.com/srid/drishti/pull/67

If I'm wrong, please re-open and reach by email (address above)!

PR #67, opened 2026-06-16
Bind loopback by default and reject SSH-option-injection hosts

srid:
**The parent server bound `0.0.0.0` with no authentication and handed an attacker-supplied host string to `ssh` as a bare positional argument** — the two ingredients of remote code execution on the machine running drishti, flagged in the security report under #43. A LAN neighbor (or any web page the operator visits, via the unauthenticated WebSocket) could reach the `__admin__` control plane, call `hosts.add` with a value like `-oProxyCommand=…`, and have `ssh` parse it as an *option* rather than a destination. This PR closes that surface at every boundary.

### The defenses

| Defect | Fix |
| --- | --- |
| Server bound all interfaces (`0.0.0.0`), unauthenticated | Bind **`127.0.0.1` (loopback) by default**; opt into wider exposure with `--bind` / `DRISHTI_BIND`, which logs a warning |
| Host string parsed by `ssh` as an option (`-oProxyCommand=…`) | Reject any host starting with `-` via a single `isValidHost`, enforced at **all three ingestion points** — admin `hosts.add`, persisted `hosts.json` on load, and CLI args |
| No `Origin` check → Cross-Site WebSocket Hijacking | Same-origin gate on the WS upgrade; `DRISHTI_ALLOWED_ORIGINS` allowlists reverse-proxy / `tailscale serve` origins |
| Tampered `hosts.json` re-seeds a payload every boot | `loadHosts` re-validates, dropping (and logging) any invalid persisted entry |

The leading-`-` rule is the load-bearing one: a real ssh destination never starts with `-` (interior dashes like `db-internal` are fine), and a no-whitespace filter alone doesn't help — `-oProxyCommand=reboot` has no whitespace, and the `$IFS` form sidesteps it.

> **On the RCE itself:** on Linux/glibc the `-oProxyCommand` payload is *incidentally* blocked today (ssh's getopt re-scans the hardcoded trailing flags and exits before connecting) — a fragile, undesigned artifact that BSD getopt on **macOS** (which drishti supports) likely does not share. So this fixes the injection at the boundary rather than relying on ssh's behavior. The loopback bind + Origin gate also independently shut the unauthenticated-attacker door (SSRF, fleet-telemetry exfiltration, persistence) regardless of the RCE.

### Adopts the upstreamed shared gate ([kolu#1386](https://github.com/juspay/kolu/pull/1386), merged)

A review surfaced that these defenses were patching a gap that lives in the shared `@kolu/surface*` libraries. Rather than fork the gate here, drishti uses the upstreamed one (now on kolu `master`):

- The CSWSH gate is **`gateWsOrigin` from `@kolu/surface/ws-origin`** — drishti's local `wsOrigin.ts` (and its tests) are deleted; the logic and tests now live upstream. drishti is WebSocket-only, so kolu's HTTP-arm gate (`gateHttpRpcOrigin`) doesn't apply.
- kolu's `surface-nix-host` now spawns `ssh … -- <host>`, closing the option-injection **structurally at the sink** for drishti's own spawns. `isValidHost` stays as the early boundary check (*defense in depth*).

The kolu pin tracks `master` (`f035cbf`); CI is green on both arches.

### Try it locally

```sh
nix run github:srid/drishti/security/harden-network-exposure-and-host-validation
```

It binds loopback; the UI is at <http://localhost:7720/>. To reproduce the old exposure deliberately: `… -- --bind 0.0.0.0` (note the warning it logs).

Addresses the report tracked in #43.

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`); kolu-gate adoption added in a follow-up commit._


srid:

Both lenses ran on the concrete diff, then cross-validated each other's recommendations.

| --- | --- | --- | --- |
| 1 | Lowy | `isValidHost` lived on the admin-wire volatility axis (`admin-surface.ts`) | Fixed in this PR |
| 2 | Hickey | `loadHosts` ran the validator over two traversals | Fixed in this PR |
| 3 | Hickey | named `isLoopbackBind` existed only to pick a log line | Fixed in this PR |
| 4 | Lowy | `wsOrigin.ts` is correctly a new encapsulated boundary | ⚠️ **No-op** |
| 5 | Hickey | `wsOrigin.ts` matches the repo's module shape; no fragmentation | ⚠️ **No-op** |

### Cross-validation (the lenses disagreed)

Hickey first judged `isValidHost`'s placement *fine*; Lowy wanted it extracted. Cross-validating surfaced the decisive detail: `__admin__` is **ssh-safe** (no leading dash), so the sentinel-exclusion is the one legitimate intersection of the two axes and must stay *inside* `isValidHost` — splitting it would fragment the "a valid host is never the routing sentinel" invariant across three call sites. The synthesis both lenses accept: **move `isValidHost` _and_ `ADMIN_HOST_SENTINEL` together** into `common/host.ts` (single import direction — no cycle, no duplication), predicate intact. That's commits `aec0338` + `35882fb`.

### Hickey rationale

The Layer-3 survey was reassuring: the diff *extends* the repo's canonical patterns rather than reinventing them — `wsOrigin.ts` mirrors the existing `log.ts` / `wakeMonitor.ts` / `archMap.ts` one-concept-per-module shape, and the `Origin` gate sits in the right position (before any RPC handler sees the socket). Two cosmetic cleanups: collapse the `loadHosts` double-pass (`4f54c1b`), and inline the log-only `isLoopbackBind` (`962cc34`).

### Lowy rationale

`admin-surface.ts`'s axis is the admin wire contract; ssh-target validity changes for unrelated reasons (a new injection vector, OpenSSH adopting `--`). The tell was that two of three importers — `hostsStore` and the CLI in `main` — are not admin clients yet had to import from `admin-surface` to validate a host. `wsOrigin.ts` is correctly a *new* boundary: its volatility (CSWSH gating) had no prior home, and its interface is stable under that axis.

> A note for the maintainer: code-police's elegance pass later suggested re-extracting `isLoopbackBind` and rewriting the `loadHosts` loop as a side-effecting `.filter`. Both were declined — they reverse the two committed Hickey findings above (a toss-up either way), and committing them would put contradictory "inline X / extract X" pairs in the history. The one conflict-free police fix (simplifying `parseAllowedOrigins`) landed as `b04f62e`.

srid:
## Evidence

Server-only security change — no intended visual change. The screenshots below are captured from a real browser hitting the new **loopback-only** bind, which exercises the new **same-origin WebSocket gate**: the gate admitted the legitimate UI, the browser connected to the parent, the parent spawned the agent, and the agent streamed live process rows back through the pipeline. A broken gate would have blocked the upgrade and left the htop table empty.

default load

*Default load — `localhost` connected, 627 processes, live system stats (CPU 80%, MEM 49%, DISK 72%); same-origin WebSocket gate passed.*

filter

*Filter applied (`bun`) — the process list narrows correctly, confirming the live data pipeline stays responsive under the new defaults.*

srid:

| sync | ✓ | 1s | forge=github, branch off master |
| research | ✓ | — | Confirmed the chain (0.0.0.0 + no-auth admin RPC + ssh bare-positional injection); mapped fixes to file:line; verified VM test curls 127.0.0.1 |
| branch | ✓ | 29s | feature branch from origin/master |
| implement | ✓ | — | isValidHost (reject leading-`-`) at wire/CLI/persisted boundaries; loopback `--bind` default + warning; CSWSH Origin gate; nix options; new tests |
| check | ✓ | 23s | `tsc --noEmit` clean (common/agent/app) |
| docs | ✓ | 33s | README + nix module updated; no stale `0.0.0.0` claims |
| fmt | ✓ | 16s | `nixpkgs-fmt` clean |
| commit | ✓ | 26s | primary feature commit pushed |
| hickey+lowy | ✓ | 13m 35s | ran + cross-validated; 3 Fix findings → commits; wsOrigin/bind/module.nix No-op |
| police | ✓ | 10m 6s | rules + fact-check clean; 1 elegance fix applied, 2 declined (reverse committed hickey findings) |
| test | ✓ | 11s | `bun test`: 136 pass, 0 fail; new behavior covered |
| create-pr | ✓ | 1m 20s | draft PR + hickey/lowy ledger comment |
| ci | ✓ | 2m 47s | odu green on HEAD (18 nodes, both arches, incl. home-manager VM test) |
| evidence | ✓ | 4m 3s | real-browser load (627 procs) proves the same-origin gate admits the UI |
| done | ✓ | 10s | summary emitted |
| **Total** | | **44m 37s** | |

**Slowest step**: `hickey+lowy` (13m 35s)

### Optimization suggestions

- **Structural review dominated (~24 min across hickey+lowy + police).** The hickey/lowy phase ran 2 reviewers + 2 cross-validators *then* applied 3 findings as separate commits; the cross-validation round-trip is the costly part and was worth it here (it resolved a genuine lens disagreement). For a follow-up of settled shape, `/do --minimal` skips docs/hickey+lowy/police/evidence.
- **police re-litigated two just-committed hickey findings** (re-extracting `isLoopbackBind`, side-effecting `.filter`), forcing a revert+adjudicate cycle. Running police *before* hickey+lowy — or scoping police's elegance pass to files hickey/lowy didn't touch — would avoid the rework.
- **`nix develop` cold-start tax** on every check/test/fmt/typecheck invocation. Running the inner loop inside a single persistent `nix develop` shell (or `direnv`) would shave the repeated devshell entry from ~5 sub-steps.
- **`research`/`implement` recorded 0s** (step-start was called after the work). Bookend `step-start` *before* the work for accurate timing next run.

Workflow completed at commit `b04f62e`.

PR #72, opened 2026-06-23
Graduate processesSnapshot to the declarative .streams.use() consumer (pulam-web Step 0)

srid:
## What

Switches `<HostView>`'s live process table from a **hand-rolled imperative stream loop** — `streamCall` + `for await` + a bespoke `AbortController` — to the **declarative `app.streams.processesSnapshot.use()` hook** plus a `reconcile` fold. The visible behavior is identical; what changes is *how* the browser consumes the stream.

## Why

This is **Step 0** of the [pulam-web plan](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/pulam-web.mdx). `.streams.use()` + Solid `reconcile` is the exact browser-consumption leg [kolu's Code tab](https://github.com/juspay/kolu) uses and the one **remote-terminals R8b** is blocked on. drishti is where that leg already runs end-to-end (browser ⇄ Node ⇄ ssh ⇄ mirror), so graduating one real consumer here — in a browser with a visible console — de-risks R8b before pulam-web exists. It's also a genuine drishti improvement: declarative replacing imperative, with the hand-rolled controller deleted.

## How it honors the design philosophy

- **Fail-fast, no fallbacks.** No new knob or degraded path. `.use()`'s `onError` surfaces stream failures to the console exactly as the old `catch` did — the caught error is reported, never swallowed into an empty state.
- **Reuse the source of truth.** Uses the framework's own `.streams.use()` hook (the canonical reactive consumer) instead of keeping a parallel hand-rolled `for await` loop. The snapshot/delta fold moves to one named home, `applyProcessesMessage`, rather than living inline.
- **Volatility boundary.** No new boundary crossed — this consumes shared `@kolu/surface` read-only (no API change), so [no drishti↔kolu gate](https://github.com/juspay/kolu/blob/master/.claude/rules/surface.md) applies.

## Key decisions

- `processesSnapshot` is a snapshot-**then-delta** union, *not* truly value-bearing, so `.use()` (which holds only the latest frame) can't render directly — the consumer must accumulate. The fold (`reconcile` a snapshot; `batch` upserts/removes for a delta) is preserved in `applyProcessesMessage`.
- The accumulation effect reads `selectedPid` via **`untrack`** so a selection change can't re-run the fold and re-apply the last delta.
- `metricHistory` (which *shared* the old `AbortController`) now gets its own; it stays imperative — Step 0 graduates only the process list.

## Tests

Adds a **hermetic test**: `implementSurface → directLink → surfaceClient → .streams.use()` driving a `reconcile` store, asserting the store re-notifies a fine-grained reader on a second delta (the reconcile-break the note calls out). `just test` runs it with `--conditions=browser` so solid-js resolves to its reactive client build — under bun's default `node` condition solid's SSR build makes `createEffect`/`createStore` a silent no-op, so the test would pass vacuously.

> [!NOTE]
> drishti's odu CI pipeline is typecheck + nix/fmt only — it doesn't run `bun test`, so this test is local/reviewer-run. Separately, `packages/common/src/surface.test.ts` fails on `master` independently of this change (pre-existing).

### DIFF
--- a/.agency/do.md
+++ b/.agency/do.md
@@ -7,7 +7,14 @@ bun run typecheck
 just fmt
 
 ## Test command
-bun test
+just test
+
+<!-- `just test` runs `bun test --conditions=browser` so solid-js resolves to
+its client build — under the default `node` condition solid's SSR build makes
+createEffect/createStore reactivity a silent no-op, so reactive client tests
+pass vacuously. Not wired into the odu CI pipeline (typecheck + nix/fmt only);
+tests are run locally / by reviewers. NOTE: surface.test.ts fails on master
+independently of this change (pre-existing). -->
 
 ## CI command
 The runner is **odu** (`github:juspay/odu`, which replaced justci — it ships
--- a/justfile
+++ b/justfile
@@ -46,6 +46,15 @@ dev host='localhost' *args: install
 typecheck: install
     {{ nix_shell }} bun run typecheck
 
+# Run the test suite. `--conditions=browser` resolves solid-js to its client
+# build (dist/solid.js): under bun's default `node` condition solid resolves to
+# its inert SSR build, where createEffect/createStore reactivity silently
+# no-ops — so a reactive client test (e.g. processesStream.test.ts) would pass
+# vacuously. The flag is safe for the node-side agent/server tests; their
+# resolution is unchanged.
+test *args: install
+    {{ nix_shell }} bun test --conditions=browser {{ args }}
+
 # Format all *.nix files (and any future biome target — drishti doesn't
 # bring biome in by default; add when JS formatting becomes a chore).
 fmt:
--- a/packages/app/src/client/App.tsx
+++ b/packages/app/src/client/App.tsx
@@ -15,6 +15,7 @@
  */
 
 import { streamCall } from "@kolu/surface/client";
+import { createSubscription } from "@kolu/surface/solid";
 import { STALE_PROCESS_CLOSE_CODE } from "@kolu/surface-app";
 import { shellCommit } from "@kolu/surface-app/lifecycle";
 import { SurfaceAppProvider, surfaceAppProbe } from "@kolu/surface-app/solid";
@@ -32,10 +33,8 @@ import {
   onCleanup,
   Show,
   type Signal,
-  untrack,
   useContext,
 } from "solid-js";
-import { createStore } from "solid-js/store";
 import {
   type ConnectionInfo,
   type ConnectionState,
@@ -48,6 +47,7 @@ import {
   type NetInterface,
   type Pid,
   type Process,
+  type ProcessesSnapshotMsg,
   type SystemInfo,
 } from "drishti-common";
 import { disconnectedMessage, STATE, withElapsed } from "./connectionColors";
@@ -65,7 +65,7 @@ import {
   pctOf,
 } from "./metrics";
 import { isActiveNic } from "./nic";
-import { applyProcessesMessage } from "./processesStream";
+import { foldProcessesMessage } from "./processesStream";
 import { coreUsageColor, processPctColor, usageBarColor } from "./usageColors";
 import {
   DEFAULT_HISTORY_WINDOW,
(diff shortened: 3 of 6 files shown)


srid:
## ⚠️ Review-gauntlet findings — HIGH regression caught; do not merge yet

This draft is **paused for a design decision**. The review gauntlet (lowy + hickey structural lenses, plus a combined fact-check / code-police / simplify pass) caught a real, reproduced **HIGH-severity regression** in the committed code. Posting findings here for later review.

### 🔴 HIGH — same-shape delta frames are silently dropped

`.streams.use()` writes each frame into a `reconcile`-backed store; the new `createEffect` reads `snapshot()` **coarsely** and copies into the `processes` store via `applyProcessesMessage`. Because `reconcile` updates nested leaves in place, two consecutive **delta** frames of the *same array shape* that differ only in a nested field (the common steady state — a hot PID's `cpuPct` ticking) don't re-trigger the effect, so the frame is **never applied**. Reproduced end-to-end through the real `defineSurface → implementSurface → directLink → surfaceClient` stack:

```
after snapshot: 10
after delta1 (cpu20): 20
after delta2 (cpu30): 20   ← frame dropped; live value frozen
```

The old imperative `for await` loop applied every frame unconditionally, so this is a behavioral regression. The hermetic test added in this PR **passes vacuously** — every frame in it changes the tracked array shape (different pids / removes), so the effect always re-fires and the bug is masked.

### Root cause — the plan's premise needs revisiting

`.streams.use()` + `reconcile` is the correct consumer for a **value-bearing** stream (each frame *is* the full value; fine-grained readers re-notify on leaf updates — exactly remote-terminals R8b's `gitStatus`). But `processesSnapshot` is a **delta-accumulate** stream, so the [pulam-web note](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/pulam-web.mdx)'s "value-bearing snapshot-then-delta" framing for it is the flawed premise. The lesson for R8b: the consumer must be value-bearing — *or* accumulated into one — and **rendered fine-grained**, never coarse-read-and-copied.

### Verified correct fix (recommended)

`createSubscription` + `reduce` (fold every frame in-loop into the full `Record<Pid, Process>` map) + render the table **fine-grained** off the subscription (`sub()?.[pid]?.cpuPct`), deleting the separate `processes` store + copy-effect. Verified against the same repro:

```
snap: 10 / d1: 20 / d2: 30   ✓ reaches 30, fine-grained reader re-notifies
```

This is the genuine value-bearing consumer R8b needs (`applyProcessesMessage`'s fold becomes the reducer). It's a larger change to perf-tuned `<HostView>` rendering (`allPids` / `visiblePids` / `selected` / `ProcessRow` / `pidComparator`), so it needs perf re-validation. Note `.streams.use()` itself has no `reduce` option, so the value-bearing path goes through `createSubscription` directly (same reactive family, lower level).

**Alternative:** revert to the imperative `for await` loop (behavior correct), which means Step 0 graduates nothing — effectively a re-scope. Recorded as the honest option.

### Structural lenses (lowy + hickey) — no defects in scope

Both lenses cleared the change as scoped (the `applyProcessesMessage` extraction, `untrack(selectedPid)`, and the metricHistory AbortController are all sound). Their one shared observation is a **next-step note, not a defect**: `metricHistory` still uses the old imperative pattern, so two stream-consumer styles now coexist in `App.tsx`; graduating it (and exporting its twin `MetricHistoryMessage`) is the seam where the follow-up step begins — deliberately out of Step 0 scope.

### Other notes

- CI not yet run: drishti's pipeline is typecheck + nix only (no test lane), so it would pass this regression green and wouldn't catch it. Will run once the approach is settled.
- `packages/common/src/surface.test.ts` fails on `master` independently of this change (pre-existing).

srid:
## Agreed — land the `createSubscription` + `reduce` + fine-grained fix

The gauntlet is right on all three counts, and this is exactly what Step 0 existed to surface. Endorsing the recommended path, with the framing for whoever lands it.

**The bug is real, and its root cause is the one that matters for R8b.** Coarse-reading a `reconcile` store misses same-shape updates (a leaf ticking in place), so the copy-effect drops the frame. That's the *same* root cause #1510 hit (reading the whole pulse object instead of the nested `seq`). **Fine-grained reads are the universal fix.**

**The premise correction is right too.** `processesSnapshot` is **delta-accumulate**, not value-bearing — each frame is a *change*, so `.streams.use()` (latest frame only) can't render it directly. I mislabeled it "value-bearing" in the pulam-web note; that's being corrected. (R8b's `gitStatus` genuinely *is* value-bearing — each frame is the full status — which is why it's the cleaner case.)

**The fix to land:**
- `createSubscription` + `reduce` — fold every frame in-loop into the full `Record<Pid, Process>` map (the old `applyProcessesMessage` becomes the reducer). `.streams.use()` has no `reduce`, so drop one level to `createSubscription` (same reactive family).
- Render the table **fine-grained** off the subscription (`sub()?.[pid]?.cpuPct`), deleting the separate `processes` store + copy-effect.
- Perf re-validate `<HostView>`'s `allPids` / `visiblePids` / `ProcessRow` / `pidComparator` path — the render shape changes.

**Fix the test, or it proves nothing.** The current hermetic test passes *vacuously* — every frame changes array shape, so the effect always re-fires. It must drive a **same-shape delta** (only a nested leaf changes, e.g. one PID's `cpuPct`) and assert a fine-grained reader re-notifies. That's the case that actually breaks.

One note on why this was worth a PR: the regression would have shipped **green** — drishti CI is typecheck + nix, no test lane, so only the review gauntlet caught it. That's exactly why graduating this in drishti *before* pulam-web/R8b earns its keep. The kolu plan (`pulam-web.mdx` + `remote-terminals` R8b) is being corrected to pin the **fine-grained-render** rule and the **same-shape-delta** test as R8b's recipe.

srid:
## ✅ Resolved — value-bearing fix landed, gauntlet clean, CI green

Following the agreed direction, the regression is fixed in `9d047c4` and re-reviewed.

### The fix (commit `9d047c4`)
- `createSubscription` + `reduce` (`foldProcessesMessage`) accumulates **every** frame in the subscription's own `for await` loop — structurally immune to coalescing, since the fold no longer depends on a reactive read re-firing.
- The table renders **fine-grained** off `processes()` (`processes()[pid].cpuPct`), so an in-place `reconcile` leaf update re-notifies exactly the changed cell.
- `applyProcessesMessage` (imperative) → pure reducer `foldProcessesMessage`; selection-on-removal clearing moved to its own effect (tracks only the selected pid's slot, so a cpu tick doesn't re-run it).
- `.streams.use()` has no `reduce`, so this drops one level to `createSubscription` (same reactive family); the `AbortController` remains only to abort the underlying stream on unmount.

### Evidence — the regression demonstrated, then fixed

The fix is behavior-*preserving* for a still frame (the table looks identical to `master`); the regression is only visible over time, so the faithful demonstration is the same-shape-delta path that broke. Reproduced end-to-end through the real `defineSurface → implementSurface → directLink → surfaceClient` stack:

```
OLD (.streams.use() + coarse copy-into-store):
  after snapshot: 10
  after delta1 (cpu20): 20
  after delta2 (cpu30): 20    ← same-shape frame DROPPED; live value frozen

NEW (createSubscription + reduce, read fine-grained):
  after snapshot: 10
  after delta1 (cpu20): 20
  after delta2 (cpu30): 30    ← every frame applied; fine-grained reader re-notifies
```

The committed hermetic test encodes exactly this (snapshot 10 → same-shape deltas 20 → 30, asserting a fine-grained reader observes 30):

```
bun test --conditions=browser packages/app/src/client/processesStream.test.ts
 4 pass · 0 fail · 12 expect() calls
```

### Re-review (fresh adversarial pass over the net diff)
Clean. Every `processes()` reader was traced — `ProcessRow`/`pidComparator` fine-grained (hot cpu reaches the cell and re-sorts under cpu-sort), `allPids`/`selected`/ppid-check coarse-by-design (key-set only), the selection-clear effect and lifecycle correct, `<For>` keyed-by-pid DOM reuse preserved. No real bugs.

**Non-blocking note:** the delta path is now O(total PIDs)/tick (`{ ...acc }` spreads the full ~470-entry map and `reconcile` walks it) vs the old O(changed). At ~470 procs / ~1 Hz this is microseconds — same order as the sort already accepted — and is the deliberate trade for the value-bearing shape. Flagging in case process counts ever grow large.

### CI — green on `9d047c4`
Full odu pipeline, **both lanes**: `18 ok · 0 failed · 0 errored · 0 skipped`.
- `x86_64-linux` (pu box `drishti-ci`) and `aarch64-darwin` (`zest`): `typecheck`, `fmt-check`, `bun-nix-fresh`, `nix`, `home-manager`, `drv-stability` all ✔. (`kolu-ci-2` was down, so the linux lane ran on the `drishti-ci` pu box; `rasam` was offline so darwin used `zest`.)

`surface.test.ts` still fails on `master` independently (pre-existing, unrelated) — and drishti CI has no test lane, so the gauntlet remains the only thing that catches client-reactivity regressions like this.

PR #80, opened 2026-06-27
perf(app): cut fleet-view CPU — aggregate, downsample, pause hidden tabs

srid:
## The problem

drishti's renderer burned **~11% of a core on the fleet view** (9 hosts) — much more than Kolu itself, and ~3.5× a single 511-row host view (~3%). Grounded in Chrome perf traces of the live instance: the cost is ~3:1 JS-over-render, spread across many tiny microtasks (zero long-tasks) — a high *rate* of decode→reconcile→DOM updates, not one heavy computation. The 511-row process table is **not** the culprit (already tamed by PID-keyed `<For>` + fine-grained reactivity).

The fleet view does work **proportional to data the user isn't consuming** — one shape in three costumes:

| Face | Before | Fix |
|---|---|---|
| **Over-subscription** | Each card opened a value stream **per CPU core per host** (144 of ~162 updates/tick @ 9 hosts) just to average them into one number | Agent folds the mean + core count into `system.cpuPct` / `system.coreCount`; the card reads those scalars and drops the per-key `cpuCores` subscription entirely |
| **Over-resolution** | Each 40px sparkline drawn from the full **900-sample** 30m ring — 27 polylines × 900 pts, ~270 KB of coordinate strings, **every 2s** | Downsample to a pixel budget (120 fleet / 240 chart) before projecting; project only while the host is live |
| **Over-frequency** | A **backgrounded** tab kept decoding every 2s frame for telemetry it couldn't show | Pause the data views after a 20s hidden grace window (sockets stay warm; history re-seeds on return) |

Plus `transition-all` dropped from the usage bars (animating a per-poll value never settles), routed through one transition-free `Bar` leaf.

## Single source of truth

The host-CPU mean was independently re-derived in **three** live places (each fleet card, the parent's metric ring, and the metric sample). It's now computed **once at the agent** — the tier that already reads per-core usage every tick — and every consumer reads that one scalar. The per-key `cpuCores` collection stays for the host drill-in, which actually renders a bar per core.

## Verification (measure-first)

- `just typecheck` ✅ · `just test` ✅ (142 pass / 0 fail, incl. new `downsample` + `captureSample`-reads-`cpuPct` tests)
- Before/after fleet-view trace (per-tick value updates 162 → ~18; sparkline points 900 → ≤120) — posting as evidence.

## Paired kolu PR

The remaining framework lever from the review — `@kolu/surface` delivering a `Collection<K,V>`'s per-tick mutations as **one batched frame** instead of N per-key frames — ships as a paired `juspay/kolu` PR (per `.claude/rules/surface.md`); this PR will pin to its HEAD and re-green CI so the batched delivery is validated downstream (it speeds drishti's host-view `cpuCores`/`networkInterfaces`).


srid:
## Evidence — measure-first (before → after)

A perf win on a monitoring dashboard is invisible on screen (the fleet renders identically), so the measurement **is** the evidence. Both traces captured with Chrome DevTools on the same machine (`pureintent`).

| | **Before** (prod build) | **After** (this branch) |
|---|---|---|
| Hosts in fleet | 9 | 3 |
| Main-thread busy | **~9.5% of a core** (1695 ms / 17.9 s) | **0.5% of a core** (103 ms / 19.1 s) |
| **Per host** | **~1.06% / host** | **~0.18% / host** |
| Sparkline points / polyline | **900** (full 30m ring) | **≤120** (capped; unit-tested) |
| DOM mutations / s | 47 (≈5.2/host) | 8 (≈2.7/host) |

**~6× less CPU per host.** The dominant, host-count-independent driver is removing the fleet card's per-core subscription fan-out — it no longer opens 16 `cpuCores` value streams per host (144 of ~162 value-updates/tick on a 9-host fleet) just to average them; it reads the agent-computed `system.cpuPct` scalar. The sparkline cap (`downsample` to 120, proven by `history.test.ts`) is a further reduction that widens as the ring fills toward its 900-sample bound.

> Note on the after-trace: the dev ring was still young (~30 samples), so part of the absolute gap is ring age. The **structural** win — no per-core fan-out — is independent of ring age, and the `downsample` cap (unit-tested at a 900-sample ring → 120) holds at a full ring.

Tests: `just typecheck` ✅ · `just test` ✅ (142 pass / 0 fail).

srid:
## Validation against final kolu HEAD (`5d69d3144`)

Pinned to the paired [juspay/kolu#1601](https://github.com/juspay/kolu/pull/1601) (opt-in `deltas` verb + its wire-gate fix). Every drishti CI recipe was run **directly** and is green against that pin:

| gate | result |
|---|---|
| `typecheck` | ✅ |
| `test` (bun) | ✅ 142 pass / 0 fail |
| `agent-boots` | ✅ (the new `drishti-common/metrics` import resolves in the agent's scoped build) |
| `fmt-check` | ✅ |
| `bun-nix-fresh` | ✅ |
| `nix-build` (app) | ✅ `drishti-built` → `drishti-client` → `drishti` |
| `drv-stability` | ✅ agent `.drv` unchanged across a client edit |

> **Heads-up:** the odu pipeline *wrapper* (`github:juspay/odu` HEAD, which this repo doesn't pin) currently can't bind its coordinator socket (`EADDRINUSE` from a clean state — a launcher bug, unrelated to this change), so the gates above were run directly rather than via `odu run`, and the GitHub status chips aren't posted. The validation is equivalent (same recipes). kolu#1601's own CI is green on both platforms.

srid:
## Review gauntlet

Ran the `/be`-review coverage on this diff (scoped to the merge-base): **lowy + hickey** structural lenses, the four **simplify** angles (reuse / simplification / efficiency / altitude), **code-police** (rules + fact-check), and an independent **codex** review.

**Clean:** codex (ran `tsc --noEmit` on common/agent/app — green; verified downsample edge cases, the visibility timer, agent empty-core handling, the `deltas` opt-in), code-police fact-check (no logic errors), and hickey (the changes are *fragmentation fixes* — collapsing the host-CPU mean that was derived in three tiers into one producer — not complecting).

**Applied:**
- **Visibility gate** (the should-fix cluster — lowy/hickey/reuse/police all flagged it): rewrote `createVisibilityGate` to compose `@solid-primitives/page-visibility` (the one SSR-safe `visibilitychange` subscription) + `@solid-primitives/scheduled`'s `debounce` (the grace window), leaving only the asymmetric "resume now / pause after grace" policy custom — and **sourced page-visibility once** in `MultiHostApp`, feeding both the gate and the becoming-visible link re-probe, so the **duplicate `visibilitychange` listener is gone**.
- **Efficiency:** `windowSlice` now short-circuits when the window already covers the whole ring (the fleet card's every-tick case → skips an O(ring) filter allocation); the agent's per-tick aggregate uses a single `Array.from(values, fn)` instead of spread-then-map.
- **Boundary doc:** corrected the `drishti-common/metrics` docstring to state the real shape (the agent is the *sole* producer of `system.cpuPct`; downstream reads the scalar) so the old client-side re-average can't be reintroduced from a stale comment.
- **Altitude:** moved the `averageCoreUsage` test next to the function in `packages/common`.

**Noted, not changed:** the `npins` kolu pin points at the `drishti-cpu` branch HEAD — the correct paired-PR interim per `.claude/rules/surface.md`; it gets repointed to kolu master once [juspay/kolu#1601](https://github.com/juspay/kolu/pull/1601) merges. A couple of stylistic nits (a `<SeriesLegend>` dedup, two trivial field-read memos) left as-is.

typecheck + tests green; paired kolu#1601 CI green on both platforms.

PR #87, opened 2026-07-05
chore: consume kolu's gray-chip fix (kolu #1687)

srid:
Paired with **[juspay/kolu#1687](https://github.com/juspay/kolu/pull/1687)** per kolu's `.claude/rules/surface.md` gate.

kolu #1687 fixes the gray Kaval chip after a fresh server start (reproduced on a pu box against a real browser):

1. **A per-key collection `get` for a not-yet-existing key now HOLDS OPEN** (subscribe-before-snapshot, yields the moment the key appears) instead of throwing `"key not found at first snapshot"` — a throw that reached the browser as a non-retriable `ORPCError` and killed the standing subscription, so a key born after boot never reached the consumer until a page reload.
2. **The re-serve's cell fold can `{ force }` past the `equals` dedup** on a rebind epoch — the ctx cell setter gains an *optional* `{ force }`.

Both are **backward-compatible** (no consumer API removed; `get`'s change is a runtime held-open semantic; the ctx setter's new param is optional), so drishti consumes them **unchanged** — this is a pin bump that validates drishti's `pumpRemoteSurface`/`implementSurface` server (its `processes`/`cpuCores`/`networkInterfaces` collections now inherit the held-open `get`) and its `createSubscription` client against the new surface source via typecheck + the nix build lane.

The pin will be bumped to kolu #1687's **final** HEAD (post review-gauntlet) before either side merges, per the surface.md re-validation clause.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -9,9 +9,9 @@
       },
       "branch": "fix/surface-boot-window-subscription",
       "submodules": false,
-      "revision": "d8d26e67624c245a986a912fb2447d1ad926d9d8",
-      "url": "https://github.com/juspay/kolu/archive/d8d26e67624c245a986a912fb2447d1ad926d9d8.tar.gz",
-      "hash": "sha256-6ta/8JhCdiDE9FZJpPW1cC/2ICYvJIjhyV/jGsv8gDc="
+      "revision": "e01cc8ea8f8e7e841d147d9af04047eb8dccf7b1",
+      "url": "https://github.com/juspay/kolu/archive/e01cc8ea8f8e7e841d147d9af04047eb8dccf7b1.tar.gz",
+      "hash": "sha256-yPX8IqNFw/UGIflaDJrT+cPxa3SPnNOxpYoyX0u2vLU="
     },
     "nixpkgs": {
       "type": "Git",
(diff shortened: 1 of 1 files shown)


srid:
CI green (10 ok · 0 failed) at the re-pinned final kolu HEAD `d8d26e676` — the surface.md gate for juspay/kolu#1687 is satisfied against the final HEAD. Ready to merge alongside #1687.

srid:
Re-ran CI on **BOTH platforms** at the re-pinned HEAD `dc96966` (kolu `e01cc8ea8`): **20 ok · 0 failed** — 10 nodes `aarch64-darwin` (@ rasam) + 10 `x86_64-linux` (@ kolu-ci-7). The earlier run was linux-only (darwin lane silently dropped when only the linux host was pinned); the surface.md gate is now satisfied on both platforms at the final HEAD.

srid:
Re-pinned to kolu **master `c97aac812`** (the squash-merge of juspay/kolu#1687) and CI-green on **both platforms**: 20 ok · 0 failed (10 aarch64-darwin @ rasam + 10 x86_64-linux @ kolu-ci-8). The surface.md gate is satisfied against the merged kolu HEAD — ready to merge.

PR #89, opened 2026-07-06
refactor: consume kolu's W4 SurfaceAppProvider accessor (juspay/kolu#1708)

srid:
The paired consumer PR for kolu's **W4 "the switch"** ([juspay/kolu#1708](https://github.com/juspay/kolu/pull/1708)) — the `.claude/rules/surface.md` gate.

## What changed in kolu
`SurfaceAppProvider.controlPlane` becomes `Accessor<ControlPlane<T>>` (the ratified S5 shape): a live host switch yields a new client, and the provider tears down the old build-identity stream and re-subscribes — a framework guarantee, so a consumer can't leak the old host's stream.

## Consumer impact
One line: `controlPlane={surfaceAppClient()}` → `controlPlane={surfaceAppClient}`. drishti's `surfaceAppClient` is already accessor-shaped (a lazy admin-client getter), so the adoption is passing the accessor itself instead of its invoked value. drishti's admin control plane is stable across the fleet, so its provider simply never trips the swap — but it compiles + runs green against the new prop type, which is the gate.

Re-pins `npins` to the kolu W4 branch HEAD (`d7ba2e9a62`). **Per the same-SHA gate, this will be re-pinned to kolu master after the kolu PR squash-merges**, and re-confirmed green there.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -9,9 +9,9 @@
       },
       "branch": "w4-the-switch",
       "submodules": false,
-      "revision": "c5f3f3014c6eae114b5524f62b603704501141fc",
-      "url": "https://github.com/juspay/kolu/archive/c5f3f3014c6eae114b5524f62b603704501141fc.tar.gz",
-      "hash": "sha256-CIFHTDbyefV+IGo0kF9LbCqaVuJKV/aYjmf/c/lDhYE="
+      "revision": "e9900fa467a92efe83f3f3384fcd32824c1eed25",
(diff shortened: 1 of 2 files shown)


srid:
Closing with juspay/kolu#1708 (its pairing partner) — superseded by the surface-family redesign; drishti's paired PR for the successor will be a substantive adoption (hostRegistry/admin-surface consuming @kolu/surface-family), not a pin.

PR #95, opened 2026-07-12
feat: surface swap as a first-class host metric

srid:
drishti watched CPU, memory, and disk — but not swap, even though it's the number that tells you *why* a host with plenty of "used" memory is still thrashing. This adds swap as a **first-class peer** of the existing three: same collection path, same alert fold, same history ring, same on-screen treatment. Nothing about it is special-cased — it rides the one shared metric vocabulary, so adding it was mostly a matter of extending `MetricKey` and letting the type system tell me every site that had to grow an entry.

## The shape

Swap is a *share of a total*, exactly like memory and disk, so it reuses the same guarded `pctOf` (`0` when total is `0`, never `NaN`). A host with swap disabled reports `0/0 → 0%` — the same "unavailable → 0" convention memory and disk already use when a source is missing, so there's no conditional hide/show branch.

| Layer | Touch point | What changed |
|---|---|---|
| **Schema** | `common/surface.ts` | `swapUsed`/`swapTotal` on the `system` cell; `swap` on `MetricSample` |
| **Agent (linux)** | `agent/proc.ts` | `parseMeminfo` now reads `SwapTotal`/`SwapFree`; `swapUsed = total − free` |
| **Agent (darwin)** | `agent/proc.ts` | new `parseSwapusage` over `sysctl vm.swapusage` (M/G → bytes) |
| **Metric math** | `common/metrics.ts` | `swapPct`; `MetricKey` gains `"swap"`; `metricPercents` projects it |
| **Alerts** | `common/alerts.ts` | one entry in `METRIC_IDS` — the fold + wire enum follow |
| **History** | `common/history.ts` (via `metricPercents`) | sampled into the ring with no change to the ring maths |
| **Client** | `client/App.tsx`, `client/metrics.ts` | `swapGb`; usage bar on the fleet card + per-host header; sky-blue history trace/legend, ordered between mem and disk |

Because `LABELS` and `SERIES_META` are exhaustive `Record<MetricKey, …>`, the compiler *forced* the label and render-metadata entries — a swap key with no on-screen treatment would not have typechecked.

## Verification

- `bun run typecheck` green across `common` / `agent` / `app`.
- `just test` — **168 pass / 0 fail**. New coverage: `parseMeminfo` swap fields (present + absent), `parseSwapusage` (M and G scaling, swapless mac, empty output), `swapPct`/`swapGb`, `metricPercents` including swap, and swap raising/clearing in the alert hysteresis fold (ordered between mem and disk).
- README metric enumerations synced (overview, fleet card, time-series charts, `system` cell reference).

## Heads-up for the reviewer

While mapping the render layer I found a **pre-existing** defect on `master`: `App.tsx:605` embeds a **raw NUL byte** as the separator in a template-literal composite key (`` `${host}\x00${id}` ``). It's committed, it makes `grep`/`ripgrep` treat the file as binary, and it's unrelated to swap — so I deliberately left it out of this PR to keep the diff scoped. Worth a follow-up to swap the raw NUL for a `\0` escape (or a visible separator).


srid:
## Evidence

Verified swap rendering live against a real host (localhost) with non-zero swap in use.

Fleet overview card showing the new SWAP usage-bar row (3.8/58.4 GB · 7%) between MEM and DISK, plus the swap legend chip in the sparkline

Per-host header showing the new "swap 3.8/58.4 GB (7%)" span between mem and disk in the top stats row

Observed swap on this host: **3.8/58.4 GB (7%)** — non-zero, so all three UI surfaces (fleet card, per-host header, history legend) render real data, not a zero-state.

srid:
## CI status

Ran the pipeline's checks locally on the **x86_64-linux** lane (the odu pipeline is typecheck + nix/fmt only):

| Check | Result |
|---|---|
| `bun run typecheck` (common / agent / app) | ✅ green |
| `just test` (`bun test --conditions=browser`) | ✅ 168 pass / 0 fail |
| `just fmt-check` (`nixpkgs-fmt --check`) | ✅ clean |
| `nix build .#default` (wrapped binary + client bundle) | ✅ built |

⚠️ The orchestrated **two-platform odu run was not launched** from this environment — there's no `~/.config/odu/hosts.json` and no `pu` CLI here, so the **aarch64-darwin lane (rasam) is unverified**. The darwin-only code path in this diff is `parseSwapusage` (`sysctl vm.swapusage`), which is covered by unit tests but has not been exercised on a real Mac. Worth a darwin CI lane before merge.

PR #113, opened 2026-07-16
Discipline the darwin cwd-enrichment child: single-flight, kill budget, backoff

### LINKED ISSUE #111: agent: process poll spawns unbounded concurrent lsof — no single-flight, CPU runaway on slow hosts
## Evidence (rasam, 2026-07-16 ~17:20)

**40 concurrent `lsof -nP -d cwd -Fpn` processes** at 90-95% CPU each, saturating all 24 cores (load 26+). Parents: drishti-agent processes (4 orphaned pre-#110 binaries — since reaped — and live ones). Mechanism: the agent's ~2s process poll spawns lsof for cwd enrichment with **no single-flight guard and no timeout**; on this host one lsof takes ~14s (an unhealthy fseventsd at 44GB RSS makes fd enumeration slow), so each tick launches another before the last finishes → ~7 concurrent lsofs per agent → N agents multiply it → slower lsofs → more overlap. Self-amplifying.

## Fix directions

1. **Single-flight**: never spawn a new lsof while one is in flight (skip the tick; the data is 2s stale at worst).
2. **Timeout + backoff**: kill an lsof exceeding ~5s and back the poll interval off while the host is slow.
3. Consider making cwd enrichment lazy/optional (it is the only reason for lsof; the process table itself comes cheaper) — on a 447-process host the enrichment is the whole cost.

Sibling of #109 (orphan agents made this N× worse — four dead generations were all still polling), but a LIVE agent alone reproduces the pileup on any slow host.

🤖 Filed by the kolu coordinator session with srid's standing authorization for drishti bug reports.

srid:
**A slow or hung `lsof` can no longer take the agent — or the host — down.** On darwin, cwd enrichment (`lsof -nP -d cwd -Fpn`, the only reason the agent forks lsof and on a big host the whole cost of the process poll) ran unbudgeted: rasam's sick fseventsd stretched one lsof to ~14s, and pre-migration agent binaries piled up **40 concurrent lsofs across 24 cores** (#111). On current master the framework poll's single-flight latch prevents the pileup but converts a *hung* lsof into **permanent collection silence** — the in-flight bit never clears, the table never updates again, and the child leaks. Fixes #111.

### The discipline (`createCwdEnricher`, owned at the reader boundary)

- **Never awaited** — the table refreshes at `ps` speed no matter what lsof is doing; rows merge the *last-landed* cwd map (stale-beats-blank: no double full-table delta churn, cwd fills one 2s tick late)
- **Single-flight** — never a second child while one is in flight, owned by the reader itself so any future caller (a second subscriber, an on-demand path) inherits the invariant free
- **Kill budget 20s + SIGKILL** — reaps the genuinely-hung child. *20s, not 5s: rasam's pathological lsof COMPLETES at ~14s — a 5s budget would kill every attempt and deliver no cwd forever on exactly the host that needs the data. Decoupling makes the budget invisible to table cadence, so generous is free.*
- **Adaptive gap** — next spawn waits 3× the last run's duration: a healthy Mac (~300ms lsof) enriches every tick exactly as before; a 14s-lsof host self-limits to one run per ~56s (≤25% of one core, vs. a pinned core forever)
- **Failure backoff** — consecutive failures (killed at budget, lsof absent) back off exponentially from 60s, capped at 5min, reset by the first success

The one hand-rolled `setInterval` left in the agent — the system/alerts tick — gets the same non-overlap law (`singleFlight`): a slow `vm_stat`/`sysctl` tick on a wedged host previously overlapped itself every 2s, the same pileup class with cheaper children.

### Grounding correction

The issue models the pre-migration poller; [the record correction on #111](https://github.com/srid/drishti/issues/111#issuecomment-4996852371) documents what current master actually does (kolu's `pollSource` already single-flights per collection) and the four residual gaps this PR closes: no timeout, 100% duty cycle under the dirty-latch, invariant not reader-owned, and the unguarded hand tick.

> **Teardown unchanged** — the #110 framework-owned exit path is untouched; at agent death at most ONE child can be in flight (single-flight); a completing child self-reaps, while the kill-budget timer dies with the agent — so a genuinely-hung child can orphan until it unwedges, bounded to one.

Pins ride in `proc.test.ts` (reader-level single-flight / never-wedges / budget-on-exec / stale-merge, red-first against the old behavior) and a fake-clock suite for the gap/backoff arithmetic, plus one real-exec contract test proving a hung child actually dies at the timeout in this runtime.

### Try it locally

```sh
nix run github:srid/drishti/poll-single-flight
```

_Generated by [`/be`](https://github.com/juspay/kolu) on Claude Code (model `claude-opus-4-8`)._

### CI provenance (final HEAD)

**Two-platform odu green on `868b5e8`** (10/10 nodes per lane: `x86_64-linux`=pu pool box, `aarch64-darwin`=sincereintent); `e0205a0` adds only `.github/workflows/nix-cache.yml` (the #114 attic fix — outside the odu DAG); **the source under test is identical**. The fresh GH-Actions `nix-cache` run on `e0205a0` is its own separate proof with the fixed workflow.


srid:
## [⚖️ Lowy ⇄ Hickey lens debate](https://kolu.dev/blog/hickey-lowy/)

✅ **Consensus** after 1 round(s) · lowy + hickey · base `17cfc86149ce`

Independent findings: lowy=7, hickey=6

### Applied (6)
- `lowy-1` Kill-budget invariant enforced per call site, not by the receptacle — commit `82c019b28`
- `lowy-2` Pruning invariant is caller-opt-in while single-flight is owned — commit `5706357bc`
- `lowy-3` Contract overclaims the recycled-pid staleness bound — commit `68f7ec3e1`
- `hickey-2` The 'every gap is capped' invariant is enforced by repetition, not structure — commit `677c24b3b`
- `hickey-3` Optional livePids makes the recycled-pid safety invariant caller-dependent — commit `5706357bc`
- `hickey-4` Enricher returns its internal mutable map, aliased across ticks — commit `5706357bc`

### Agreed — no change (7)
- `lowy-4` Non-overlap law spelled three ways; canonical receptacle exists (packages/agent/src/main.ts:93)
- `lowy-5` Generic child-discipline electricity fused with lsof specifics (packages/agent/src/proc.ts:707)
- `lowy-6` Domain-agnostic utility exported from the most volatile layer (packages/agent/src/main.ts:93)
- `lowy-7` Contract layer's docs name the implementation module (packages/common/src/surface.ts:46)
- `hickey-1` EXEC_BUDGET_MS complects kill-budget policy with retry-cadence policy (packages/agent/src/proc.ts:747)
- `hickey-5` Generic concurrency combinator housed in the CLI entrypoint module (packages/agent/src/main.ts:93)
- `hickey-6` Enrichment-discipline constants fragmented into three prose copies (packages/common/src/surface.ts:44)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 rounds (2 fix commits) · reviewer effort: xhigh · base: merge-base 17cfc86

| Round | Commit | Description |
|---|---|---|
| 1 | b6b6673 | fix(codex): round 1 — execFile boundary, landing intersect, honest staleness docs |
| 2 | 5a5aa16 | fix(codex): round 2 — F2 for real: per-run MONOTONE eligible set |

**Legend** — findings codex raised:

- **F1** — The budget is implemented with `child_process.exec`, which launches a shell; timeout/SIGKILL targets that spawned shell and does not guarantee that the actual `ps`, `vm_stat`, `sysctl`, `netstat`, or `lsof` process is terminated
- **F2** — A slow in-flight `lsof` result replaces `lastMap` wholesale after intervening polls may already have pruned dead PIDs
- **F3** — The public contract says recycled-PID staleness is bounded by the five-minute backoff ceiling and that cwd fills one poll tick late
- **F4** — The max-buffer rationale says 16 MiB is bounded to one child at a time, but the guards are per reader and `readSystem` deliberately runs `vm_stat` and `sysctl` concurrently; independent process, network, system, and enrichment polls can overlap as well

srid:
## [👮 Code-police](https://agency.srid.ca/) + architecture checks

**Architecture-first-principles (C1–C6, adversarially verified, ran first):** 3 confirmed findings, all fixed in 058cd16 — a hung `ps` still wedged the processes collection (the #111 wedge one exec to the left; every darwin child now rides one kill budget), the new `singleFlight` had converted a hung `vm_stat` into a permanent silent freeze (same budget fix), and the success-path gap was wall-clock and uncapped (monotonic clock + 5min cap). 22 minors dispositioned (cheap ones fixed; graduation/migration notes recorded). No project overlay exists — checks ran generic.

**Police Pass 1 (rules):** one violation — `no-silent-error-swallowing`: the enricher's failure backoff left **zero diagnostic trace** for an operator debugging a blank/stale cwd column. Fixed in b8eb635: one stderr line per failure episode (with the cause) + one on recovery; the retry cadence stays quiet.

**Police Pass 2 (fact-check):** two findings, both fixed —
- `statfs("/")` was the one timeout-less await under the settlement guards (a D-state root fs froze the system cell + alerts forever). Fixed in b8eb635: single-flight probe + last-known-value cache; the tick never awaits the syscall.
- The kill budget turns a hung-at-boot `ps` into a **seed-read rejection**, which the framework's poll contract makes permanently fatal to that collection — and nothing observed `runtime.done`, so a live agent could silently serve a dead table for its whole life. Fixed in 868b5e8: `serveAgent` observes `runtime.done` and exits loud (agent death surfaces in the parent's connection cell); deliberately *not* fixed by making reads total, which would render a healthy-looking empty table over a broken platform.

**Simplify pass** (e014412): framework `monotonicNow` reuse, single-flight latch ⊕ eligible-set merged into one value, per-tick snapshot return (structural no-caching), test/doc dedup. Skips recorded with reasons in the commit message.

Elegance pass skipped (`--no-elegance` — the standalone simplify track ran).

PR #125, opened 2026-07-27
Show a failed host's evidence, read off the record that survives

srid:
Pairs **[juspay/kolu#2022](https://github.com/juspay/kolu/pull/2022)**, pinned at kolu `c432080fae4433f660eb4e84e415b6772d288f05` (branch `surface-failure-evidence`).

## What moved upstream

Failure **evidence** became a framework-minted co-product of the failure record in `@kolu/surface-map` / `@kolu/surface-remote`:

- `EntryStatus.failed` now requires `evidence: FailureEvidence` (`readonly {source: "local" | "remote"; line: string}[]`) beside its domain `failure`, and `entryStatusSchema` refuses a failed status without it — reason-without-evidence is unspellable in the type *and* on the wire.
- `serveHostMap` staples the session's retained log tail on at the same seam it classifies the reason.
- `floorOnLiveness` is unchanged: it keeps `failure` (now carrying evidence) and drops only `connection`, so the evidence survives a dead link by construction.

## What compiled unchanged

Everything on the server, and every existing read of the failure. `admin-router.ts`'s `serveHostMap` call publishes evidence with **zero** code changes — the stapling happens inside the framework — and `entryStatusTone.ts`'s `status.failure.reason` reads are untouched. The single compile break in the whole repo was one test fixture that constructed a `failed` arm without `evidence`, which is exactly the error the tightened type is for.

## What we adopted, and why it is not cosmetic

`FailedCard` took its "Connection log" from the **live** `connection.log`, and `ConnectingOverlay` narrowed the failed arm out of `connection.phase`. Both are the wrong source, for the same reason: `connectSurfaceMap`'s liveness floor **drops** `connection` over a dead admin link while **keeping** the failure. So a host that failed while our link to the browser was down rendered as "connecting…", with the actual ssh / `nix copy` output floored away — [juspay/kolu#2007](https://github.com/juspay/kolu/issues/2007)'s defect, reproduced on drishti's side.

Both now read the entry's failure record through one new seam:

- `packages/app/src/client/entryStatusTone.ts:94` — `failureRecord(status)` returns `{reason, evidence}` or `null`, beside the `dotClass`/`statusTitle` the chip already reads, so the failure page and the dot cannot disagree about whether a host is failed.
- `packages/app/src/client/App.tsx:1247` — `HostView` derives the standing failure off `entryState()` and passes it to `ConnectingOverlay`, which renders `FailedCard` from it instead of from the live payload.
- `packages/app/src/client/App.tsx:1682` — `FailedCard` renders the last 8 evidence lines, monospace, only when non-empty, each keeping its `source` tag. Whether drishti's own parent said a line or the far end did is often the diagnosis, and the field carries that rather than an in-band prefix. Its `reason` prop tightened from `string | null` to `string` on the way through: the framework has no fabricated fallback cause, so a failed entry always carries a real one.

The dot's `title` tooltip is unchanged — one line is right for a tooltip, and a log does not belong in one.

One small pin covers the new seam (`entryStatusTone.test.ts` — the reason arrives *with* its evidence, and an up host yields nothing).

## Verification

Typecheck, 191 unit tests, and full odu CI — `nix`, `typecheck`, `fmt-check`, `bun-nix-fresh`, `home-manager`, `drv-stability`, `agent-boots` — **20/20 green on both `x86_64-linux` and `aarch64-darwin`**.

## ODU-IMPACT VERDICT: `none`

Grounded by a grep of [`juspay/odu`](https://github.com/juspay/odu) at its pinned kolu SHA (`3ff256f0`): odu hydrates `@kolu/surface-map` in nix (`nix/overlay.nix`, `nix/env.nix`) but imports **nothing** from it in TypeScript — no `EntryStatus`, no `EntryState`, no `serveHostMap`, no `EntryFault`. Its `@kolu/surface-remote` use is `makeSession` / `sshConnector` / `SessionState` (`src/coordinator/lane.ts`, `lease.ts`, `runnerFlake.ts`), and kolu#2022 does not touch `session.ts` or `SessionState` — evidence is the *existing* `SessionState.log` passed through at the map seam. Nothing to break and nothing to adopt at odu's next bump.

## Re-pin to follow

kolu#2022's review gauntlet is deliberately held right now, so this pin is a **branch** SHA. Per kolu's `.claude/rules/surface.md`, a drishti green pinned to a pre-gauntlet SHA is stale: **this PR will be re-pinned to kolu's final post-gauntlet HEAD, re-merged with `origin/master`, and re-validated green before merge.**

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -36,6 +36,7 @@
         "@tailwindcss/cli": "^4.1.0",
         "babel-preset-solid": "^1.9.0",
         "cleye": "^2.3.0",
+        "dequal": "^2.0.3",
         "drishti-common": "workspace:*",
         "hono": "^4.12.16",
         "neverthrow": "^8.2.0",
@@ -274,6 +275,8 @@
     "csstype": ["csstype@3.2.3", "", {}, "sha512-z1HGKcYy2xA8AGQfwrn0PAy+PB7X/GSj3UVJW9qKyn43xWa+gl5nXmU4qqLMRzWVLFC8KusUX8T/0kCiOYpAIQ=="],
 
     "debug": ["debug@4.4.3", "", { "dependencies": { "ms": "^2.1.3" } }, "sha512-RGwwWnwQvkVfavKVt22FGLw+xYSdzARwm0ru6DhTVA3umU5hZc28V3kO4stgYryrTlLpuvgI9GiijltAjNbcqA=="],
+
+    "dequal": ["dequal@2.0.3", "", {}, "sha512-0je+qPKHEMohvfRTCEo3CrPG6cAzAYgmzKyxRiYSSDkS6eGJdyVJm7WaYA5ECaAD9wLB2T4EEeymA5aFVcYXCA=="],
 
     "detect-libc": ["detect-libc@2.1.2", "", {}, "sha512-Btj2BOOO83o3WyH59e8MgXsxEQVcarkUOpEYrubB0urwnN10yQ364rsiByU11nZlqWYZm05i/of7io4mzihBtQ=="],
 
--- a/bun.nix
+++ b/bun.nix
@@ -444,6 +444,10 @@
     url = "https://registry.npmjs.org/debug/-/debug-4.4.3.tgz";
     hash = "sha512-RGwwWnwQvkVfavKVt22FGLw+xYSdzARwm0ru6DhTVA3umU5hZc28V3kO4stgYryrTlLpuvgI9GiijltAjNbcqA==";
   };
+  "dequal@2.0.3" = fetchurl {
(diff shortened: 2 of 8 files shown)


srid:
Follow-up: this merged carrying the pin at kolu `928d9bced`, the head of `surface-failure-evidence`. kolu#2022 was squash-merged, so that commit is not an ancestor of kolu master, and the branch has since been deleted upstream — the pin names a commit no ref reaches. The archive tarball still returns 200 so nothing is broken today, but it is GC-able.

#126 repairs it: same content (`git diff 928d9bced 0964acb3a -- packages/surface*` is empty), repointed at the #2022 merge commit `0964acb3a` with the pin's branch field on `master`. Deliberately not current master, which would pull in untested kolu#2018.

PR #128, opened 2026-07-29
Make the agent a durable daemon (UW3)

srid:
**Metric history and alert state now survive reconnects, parent deploys, and agent upgrades.** The agent is a durable surface daemon under `~/.local/state/drishti/`; the parent still dials `drishti-agent --stdio`, but that flag is a front that adopts-or-spawns the gate-held daemon and relays bytes.

### One binary, two modes

```
ssh … drishti-agent --stdio  →  frontDaemonOverStdio (adopt / spawn + relay)
drishti-agent                →  daemonMain on unix socket
                                idle-exit after 60 min with no connections
```

Files: `drishti.pid`, `drishti.sock`, `history.ring.json` via `daemonHome({ app: "drishti", placement: "state" })` — never `$XDG_RUNTIME_DIR`.

### History ring on disk

The ring moves into the agent (sampled every poll tick, flushed every 30s and on drain). Atomic temp+rename. Unknown `v` or garbage → typed `unavailable` chart state (file left alone or moved aside — never silent empty).

### Identity + convergence

Agent serves `controlCoreFragment` at contract `"1.0"`. Parent policy:

| Axis | Arm |
| --- | --- |
| capability | drainable |
| contract skew | drain-newer-else-refuse |
| build mismatch | drain-and-replace |
| drain budget | maxAttempts: 2, onGiveUp: adopt-stale |

Only the policy object and link-state projection are drishti-written — probe + decision table live in `@kolu/surface-daemon-supervisor`. Host sessions are `DaemonSession`s (`convergence()` / `renew()`).

### Fleet BUILD_ID freeze (U6.1)

Agent identity is the **inner wrapper path hash**. The agent fileset is a positive projection only: `packages/agent` + `packages/common` + `agent.package.json` (exact workspace manifest) + `agent.lock` / `agent.bun.nix` / `agent.bunfig.toml` + single-file `scripts/hydrate-kolu-packages.sh` + `tsconfig.base.json`. Root `package.json` and the rest of `scripts/` are **not** inputs.

> **one-time identity migration — input-set isolation; no daemon-capable release predates it.**

Frozen map after kolu master re-pin (`agentBuildIdsJson` — rotates when kolu/osfacts inputs change, not on app-only edits):

```json
{"aarch64-darwin":"d6db39c578df4077aeb1a06cc251de5ba4810f46acdf28f1c728d359b7f37ef1","aarch64-linux":"10ef8b6aec9139a0ffde3def070359f16b0c66c45804035167cab7bddb639310","x86_64-linux":"a3ba2951f343fbd00288e873069e2bf575de0ee6fe29ea0a9c26cd4a97714705"}
```

`just ci::drv-stability` probes (full-map byte-identical unless noted): client-ts, real app-only devDep + lock/`bun.nix` regen, root-manifest-only, app-only script (`gen-pwa-icons.ts`); agent-touch rotates.

### Daemon glance (kaval shape)

Host entries always show a **subtle daemon glyph** (no status text) that opens the daemon dialog. The **loud status chip** is anomaly-only — quiet when healthy so it does not duplicate the connection label.

### Tests

- planted `v+1` and truncated ring dispositions
- shared-artifact registry + watchdog
- synthetic mixed-build drain (bootstrap: no previous daemon tag yet) re-reads `history.ring.json`
- policy shape pins
- Solid glance confinement + quiet-when-healthy loud chip + always-present glyph
- agent BUILD_ID isolation from app lock/bunfig/test-preload churn
- both-or-neither `readBakedIdentity` plants under kolu UW5

Kolu pin: `juspay/kolu` **master** @ `5ec6ad09` (UW5 / PR 2047 merged).

### Try it locally

```sh
nix run github:srid/drishti/uw3
```

_Generated by [`/be`](https://github.com/srid/agency) on Grok Build (model `grok-4.5`)._


srid:

| # | Lens   | Finding | Disposition |
|---|--------|---------|-------------|
| 1 | Hickey | History view shattered across three layers | Fixed in this PR |
| 2 | Hickey | Contract version triplicated as `"1.0"` | Fixed in this PR |
| 3 | Hickey | Convergence policy constructor duplicated | Fixed in this PR (version shared; synthetic policy stays package-edge local) |
| 4 | Hickey | isIdle counts metricHistory, not connections | Fixed in this PR (renamed + documented lease) |
| 5 | Hickey | Artifact test reinvents gate/socket with wrong ids | Fixed in this PR |
| 6 | Hickey | historySurface dead alias | Fixed in this PR |
| 7 | Hickey | hostRegistry buildEntry braids admit lifecycle | Fixed in this PR |
| 8 | Hickey | awaitExit oracle reimplemented in mixed-build | ⚠️ **No-op** (agent ↛ app import edge; test stays self-contained) |
| 9 | Hickey | buildAgentRuntime null-then-assign | Fixed in this PR |
| 10 | Lowy | App surface version dual-homed | Fixed in this PR |
| 11 | Lowy | MetricHistory fold sprayed across tiers | Fixed in this PR |
| 12 | Lowy | Dead historySurface parallel receptacle | Fixed in this PR |

### Hickey rationale
Structural complecting centered on history disposition (three parallel representations of one sum type), unenforced multi-site contract version, and an admit factory that braided too many concerns into `buildEntry`. Simplifications: `HistoryView` + `foldHistoryView`, single `AGENT_SURFACE_VERSION`, extracted admit helpers, honest isIdle naming.

### Lowy rationale
Volatility axes: surface contract version must be one symbol shared by producer and admit policy; metric-history message fold is one state machine and belongs in the pure history module; dead `historySurface` was an expendable leftover receptacle after metricHistory joined the agent surface.

_Generated during `/be` gauntlet on Grok Build._

srid:
## Claude ⇄ Grok debate

✅ Consensus in 3 round(s) · base `origin/master`

| Round | Commit | Description |
|---|---|---|
| 1 | 7c61428 | fix(agent): address Claude debate round-1 findings |
| 2 | 6bd2e59 | fix(agent): withhold persist when ring file is unreadable (F13) |

**Legend** — findings Claude raised:

- **F1** — Once historyView loads as unavailable, the daemon never recovers: the tick gate (main.ts:404) stops sampling and flushRing (main.ts:227) stops persisting, for the entire life of a durable daemon that may run for weeks
- **F2** — The parent's expected build identity is the single DRISHTI_AGENT_BUILD_ID baked in default.nix — the hash of the agent built for the PARENT's system
- **F3** — moveCorruptAside swallows every rename failure silently: the catch checks for ENOENT and returns, and every other error code also falls off the end of the function with no log
- **F4** — A non-ENOENT read error (EACCES, but also transient EIO/EMFILE/ENFILE) is classified as 'corrupt' AND triggers moveCorruptAside — renaming a possibly perfectly healthy ring file aside (rename needs only directory perms, so it can succeed where read failed)
- **F5** — Contradictory comments about budget scoping: lines 344-348 say 'ONE policy + budget for the whole pool … mint once, share across hosts' admit hooks … one shared budget is fine', then lines 351-353 and buildEntry (line 358-359) say and do the opposite — a budget is minted PER host entry 'so one host's flap doesn't exhaust another's budget'
- **F6** — processExit = conn.closed.then(...) has no rejection handler on any path: if conn.closed ever rejects, processExit becomes a rejected promise with no catch (awaitExitViaProcessOracle attaches only .then, and if no drain is ever attempted nothing attaches at all) — an unhandled rejection that can crash under strict rejection settings
- **F7** — Snapshot-then-subscribe gap in the metricHistory stream source (agent main.ts:296-315, and the same shape at packages/app/src/server/router.ts:178-195): the snapshot is copied at first pull, but historyBus.subscribe(signal) is only called after the consumer pulls past the first yield
- **F8** — The ~20-line 'yield current HistoryView as snapshot-or-unavailable, then forward the bus' async generator is duplicated byte-for-byte in spirit between the agent (packages/agent/src/main.ts:296-315) and the parent (router.ts:178-195)
- **F9** — reExecAsDetachedDaemon points the daemon's stderr at home.file('agent.stderr.log') with no visible rotation or bound
- **F10** — serveAgent's `serve` parameter now defaults to `async () => {}` — a production-named export whose default silently serves nothing and returns immediately
- **F11** — Explicit `-h`/`--help` routes to usage(), which exits 1 (main.ts:119)
- **F12** — saveHistoryRing writes a pid+timestamp-named temp file then renames; if renameSync throws (target dir perms changed, cross-device surprise), the temp file is orphaned
- **F13** — Regression from the F1+F4 interaction: for a non-ENOENT read error, loadHistoryRing now (correctly, per F4) leaves the file in place and returns unavailable('corrupt') — but buildAgentRuntime's corrupt branch assumes 'file was moved aside — free path' and resumes active persistence

PR #132, opened 2026-08-03
Take kolu's Effect 4 migration: Effect RPC end to end, and a drain that can say what happened

srid:
Pairs with **juspay/kolu#2101**.

kolu's `effect` branch replaces oRPC + zod with Effect RPC and Effect Schema across the whole `@kolu/surface*` stack. drishti vendors those sources straight out of the Nix store, so it does not get to opt in later — it takes the same migration or it stops compiling. This is that migration: the pin, the framework re-wiring, and drishti's own zod → Effect Schema conversion, across a stack of staged commits.

> ✅ **juspay/kolu#2101 has LANDED** — squash commit `05484f0fb` — so the pin stops chasing a branch. It now tracks kolu **`master` at `2104ddea9`**, which carries the merge plus the master-side work that never existed on `effect` (release 2.2.0, the changelog rules, the `/release` CI-gate delegation, the operations page). The osfacts-client adoption (opened as srid/drishti#133) is folded into this branch — this is the one PR for this repo.
>
> **CI at the master pin: two-platform FULL GREEN** — **24 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled** across `x86_64-linux` and `aarch64-darwin`, all **24 GitHub statuses `success`** at `05e3482`, **no rerun and no flake**. The heavy daemon-gated lane (`ci::test-daemon@x86_64-linux`, `KOLU_DAEMON_TESTS=1`) ran **389 pass · 0 fail across 56 files** in 78s — the epoch-gate e2e and the G-round fault-wiring pins included. Receipts: [PR checks](https://github.com/srid/drishti/pull/132/checks) · [commit statuses at `05e3482`](https://github.com/srid/drishti/commit/05e3482a104836ab6e4c17e7d9be4f0b4e8e227d).

## Round: the mute death (kolu#2101 G-round, deploy #2)

The second deploy incident had a different signature from the first. The epoch gate stopped a client attaching to a peer that could not speak; this one is about a process that keeps **answering** after its insides have died — alive, gate held, socket accepting, runtime dead — and about every channel that was supposed to report it being either absent, silent, or filtered away.

kolu's fix is framework-wide (callback containment, cell-local poll seeds, a typed `onFault`, a daemon fault-exit arm, an explicit frame cap). drishti had two instances of the class and one near-miss. This section is the audit, with the disposition for each.

### The G-round audit

| Delta | drishti's exposure | Disposition |
| --- | --- | --- |
| `DaemonSpec.faultSignal` / `armRuntimeFaultExit` — observing `runtime.done` with log-and-continue is now contractually wrong | **Exposed, differently than expected.** drishti already treated a fault as fatal — but via a bare `process.exit(1)` *inside the runtime builder* | **Fixed.** Rides `armRuntimeFaultExit` → `faultSignal` |
| Same, parent side | **Exposed.** The per-host bridge's `implementSurface(...).done` was never read at all | **Fixed**, loud but deliberately non-fatal |
| `MirrorRemoteSurfaceOptions.onFault` (typed `MirrorFault`) | One call site, `pumpRemoteSurface`, passing `log` only | **Adopted.** Faults now on their own channel |
| `mirror.done` rejects on member fault / settles promptly | Same call site; the promise is `void`ed | **No change** — residual restated below |
| Reactor poll cells: T+0 seed failure is now cell-local | Five poll sources, all with throwing reads | **No change needed** — and a latent bug disappears |
| `websocketLink` throwing `url()`/`connect()` thunk | **None** — zero `websocketLink` calls | **N/A**, argued below |
| `frameLimit` / `rpcSerializationLayer` | Every serve site is a kolu primitive | **Inherited**, no drishti edit |
| Framework callback containment | Consumer callbacks throughout | **Free** — strictly a safety gain |

### The agent daemon exited on a fault — around the shutdown spine, not through it

`buildAgentRuntime` carried `void built.done.catch(… process.exit(1))`. The **verdict** was already right, which is exactly why this was easy to miss: a runtime fault is structural death, and drishti said so. The **mechanism** was wrong. A bare exit from inside the runtime builder skips `daemonMain`'s teardown, so the unix socket and the pid gate were never released and the history ring was never flushed. The successor then met a gate naming a dead pid, and the ring had lost every sample since the last 30 s persist.

It now rides `armRuntimeFaultExit` → `DaemonSpec.faultSignal`: ordered teardown, the ring flushed as **last rites** (drishti's analogue of padi's final-session capture — the ring is the one piece of agent state a respawn cannot reconstruct), and an exit that `daemonExitCode` scores `runtime-fault`, which is the supervisor's only channel for *that was a crash, not a stop*.

The observation moved out of `runtime.ts` entirely. A library that builds a runtime has no business killing a process it does not own — and the test path (`serveAgent`) has no daemon to exit at all. `done` is handed out; the daemon binary decides.

### The parent dropped its bridge's fault on the floor

`router.ts`'s per-host `implementSurface(mirroredAgentSurface, …)` runtime had its `done` **never read**. That is the silent half of the same class, parent-side: the bridge's reactive wiring dies, the host's projection stops updating, and nothing anywhere says so.

It is observed now. Deliberately **not** fatal, and the asymmetry is the point: the parent serves every host, and one bridge dying is not grounds to take the other hosts' canvases down with it. So the disposition is a loud, attributable line, and the residual is named in the code rather than implied — a host whose data is frozen behind a chip the session still calls connected. What the UI *should* say there is a product call, not something to smuggle into a pin bump. It joins the `pumpRemoteSurface` residual already flagged in this PR; they are the same open question seen from two angles.

### Faults stopped sharing a channel with chatter

The pump passed `log` and nothing else, so a member's stream dying read as narration.

Worth being precise about what drishti was and was not exposed to. kolu's incident was **mute**: its fault line went to `log.debug`, which production filters out, so a dead mirror produced not one line. drishti's `Logger` is a bare `(line: string) => void` with no levels and no filtering — every line reaches stderr — so the disappearing-log half **could not happen here**. What drishti lacked was the *distinction*.

`onFault` now carries a `FAULT`-prefixed line with the error's own stack, and reports `scope` verbatim: `member` (the mirror can never re-subscribe; the pump's promise rejects) and `key` (one key died, the host's mirror survives by design) must not look alike, or a survivable fault reads like a dead host.

### Two audits that needed no code, and are not "no-ops"

**Poll cells — and a latent boot-death that disappears.** drishti has five reactor poll sources (`processes`, `unclaimedListeners`, `sourceErrors`, `cpuCores`, `networkInterfaces`), every one with a read that genuinely throws — `readProcesses()` and friends raise `OsfactsSourceError` when a facet is missing. Nothing in drishti relied on a T+0 seed failure being fatal: the recurring `tick` already swallows read errors and keeps serving, and the one seed-is-fatal path (`readSystem()` at construction) is a plain `await`, not a reactor cell.

The interesting part is the direction of the change. *Before* this release a T+0 seed throw was fatal to `runtime.done` — which drishti's old code turned into `process.exit(1)`. So a single transient osfacts hiccup at daemon boot would have killed the daemon, and with the fault-exit wiring above it would now do so *properly*, which is worse, not better. The G-round makes that seed failure cell-local (hold the default, retry next tick), so the combination is correct: transients no longer reach the fault channel, and only structural death does. drishti gets a latent boot-fragility removed for free, and the fault-exit stays meaningful because it can no longer fire on a hiccup.

**websocketLink — genuinely not applicable.** Zero calls. drishti's browser client dials through `connectSurfaces` (`@kolu/surface-app/solid`), so the framework owns the `url()`/`connect()` thunks and the re-dial classification lands entirely inside kolu. drishti's one URL thunk (`wsBase()`) is passed *to* `connectSurfaces` and is now covered by the framework's own containment.

### The frame cap: inherited, with one residual measured rather than asserted

Every drishti serve site is a kolu primitive — `daemonMain` → `serveOverUnixSocket`, `surfaceAppLayer`, `serveSurfaceSocket`, `serveHostMap` — and each applies `rpcSerializationLayer` internally now. So the 16 MiB cap is explicit on drishti's wire without a drishti edit, and there is no silent-Effect-default left anywhere in the path.

The payload audit found one bounded and one unbounded shape.

`metricHistory`'s snapshot frame is the largest **bounded** one, and it is pinned as a number rather than a claim: 30 min retention ÷ 2 s poll = **900 samples**, generously padded to 160 B each ≈ **144 KB** — over 100× under the cap, asserted in `runtimeFaultWiring.test.ts` so a future retention or cadence change that erodes the headroom fails there instead of on a user's socket.

The **residual** is the `processes` collection. Nothing caps the process *count*, and each row carries an unbounded `listeners` array; only the per-row `command` and `cwd` strings are truncated (200 chars). At ~400–600 B/row that is roughly 30k processes to reach the cap — implausible on an ordinary host, less so on a container-dense box, and the listeners array is a multiplier that lowers the threshold on a socket-heavy one. It is flagged rather than fixed: capping it means silently dropping processes from a process monitor, which is a product decision about what the user is owed, and inventing that in a pin bump is how a monitoring tool quietly starts lying. The cap's consequence is worth stating plainly — an oversized frame closes the whole socket (1009), so it would take every unrelated subscription on that tab with it, not just the process list.

### Falsified

`packages/agent/src/runtimeFaultWiring.test.ts` pins both halves of the fault wiring (the exit is gone from the builder, `done` is still handed out; `main.ts` arms off `runtime.done`, passes `faultSignal`, and uses `flushRing` as last rites) and the frame-limit headroom. Reverted to the bare `process.exit`, the two wiring pins fail and the headroom pin correctly does not, being independent of it.

These are source pins, and deliberately so: the wrong version of this code *also compiles* and *also handles* the fault — the old `process.exit(1)` did both. The defect was never a missing handler, it was a handler that skipped teardown, which is invisible to `tsc` and visible only in the shape of the call. One wrinkle worth recording: the pins strip comments before matching, because the rationale comments quote the very `process.exit` they exist to explain — a test that punishes documenting the fix is worse than no test.
## Round: the stdio epoch gate, and drishti's fleet arm (kolu#2101 F5b)

kolu's production incident — every remote host wedged in a permanent connect loop — turned out to be a **framework**-level hole, so the fix is framework-level and drishti takes it the same way it takes everything else in this PR: by re-pinning. `stdioLink` now REQUIRES an un-forgeable `StdioReadinessProof`, minted only by `awaitStdioReadiness` off a peer's own `ready` banner; `duplexWireLink` is no longer exported; `serveOverStdio` greets at boot; `sshConnector` reads the banner before it will build a client. The invariant is *no pinger before a proven epoch*.

kolu's review comment asks this PR to answer one question directly (F5b): **does drishti's fleet arm have the same class?**

It does, and it had more of it than kolu's remote arm did.

`drishti-agent --stdio` handed `frontDaemonOverStdio` a socket path and spliced into whatever accepted — no gate read, no identity probe, no epoch check. kolu at least ran the full convergence kit on its *local* arm; drishti's only supervisor of an agent daemon was the parent's `convergeAdmit`, an ssh hop away from the gate file, the pid table and the signals it would need to act on. And drishti's agent daemon is **resident**: it idle-exits after 60 minutes, so it routinely outlives a deploy. A previous-epoch drishti agent is not a rare race — it is the ordinary state of every host between a deploy and the next idle timeout, which is exactly the population the incident wedged.

### The F5b audit — every stdio site in the repo

Swept for `stdioLink`, `duplexWireLink`, `socketDuplexLink`, `serveOverStdio`, `frontDaemonOverStdio`, `unixSocketLink`, `sshConnector` and any hand-rolled splice across all of `packages/`.

| Site | Class | Action |
| --- | --- | --- |
| `agent/src/main.ts` — `drishti-agent --stdio` | **Daemon front over a RESIDENT daemon** — the incident class | Converge-before-relay pre-step + readiness banner |
| `agent/src/fixtures/{high,low}ContractMain.ts` (via `fixtures/skewFront.ts`) | Daemon front (contract-skew fixtures, same shape) | Same pre-step, shared verbatim — see below |
| `agent/src/dialDaemon.testlib.ts` — `dialOverStdio` | Direct `stdioLink` construction | Mints a real proof via `awaitStdioReadiness` on the front's own stream |
| `agent/src/convergeFront.ts` — `socketDuplexLink` (new) | Local unix rendezvous | Documented residual; its epoch safety is the very thing the converge establishes |
| `agent/src/dialDaemon.testlib.ts` — `unixSocketLink` | Local unix rendezvous | Unchanged; no proof owed |
| `app/src/server/hostRegistry.ts` — `sshConnector` ×3 | Dialer | No source change; it now awaits the banner. The boot barrier around it needed one (see below) |
| `serveOverStdio` | — | **Not used.** drishti's daemon serves a unix socket; nothing here is an ephemeral stdio agent |
| `duplexWireLink` | — | **Not used.** The removed export costs this repo nothing |

Two `.pipe(` hits in the sweep are Effect pipes, not stream splices.

### The front converges, then greets, then relays

The pre-step runs the real kit on the box where the facts live: the `probeDaemonIdentity` **factory** (the raw dial with the byte tap and the silence deadline — the only thing that can *see* a previous-epoch peer, unlike the assembly-only `probeDaemonIdentityFrom` the parent's admit uses), gate corroboration against the pid table, takeover via `reapHolder`, and `reExecAsDetachedDaemon` as the driver so a replacement is an agent of *this* closure. Converged, it greets `ready` and splices. Unconverged, it writes a `refused` banner carrying the typed anomaly and exits non-zero **without relaying**.

A side effect worth naming: drishti's `unspeakable-protocol` daemon-status arm has been **dead-but-typed** since it was written. Its only producer is the framework's endpoint-arm `converge`, which drishti never called — every path went through `convergeAdmit`, which structurally cannot emit it. This pre-step is its first producer, so the arm now means something.

### The front's policy is deliberately NOT the parent's

This is the one real design decision in the round, and it is worth stating plainly because the obvious move is wrong.

The front's question is the **epoch** question, and only that: *does a daemon that speaks this wire at all hold this rendezvous?* That is what the banner attests, and it is what the incident turned on — a previous-epoch peer accepts the splice and then says nothing, so no amount of in-epoch reasoning can even begin.

Every **in-epoch** verdict — contract skew, build mismatch, the drain budget, the standing anomaly, the degraded chip, the renew affordance — stays with the parent's `convergeAdmit`. Those already work end to end and have been tested since W6, and the parent is the only one of the two supervisors with a UI.

So the front's policy is `not-drainable`, with `refuse` on contract skew and `nudge-human` on build mismatch — both meaning *leave the survivor standing and report it*. `not-drainable` is the structural statement: with no drain arms and no budget spellable, a front **cannot** quietly grow into a second adjudicator of skew.

The discriminator is a `skew-refused` outcome, which means the resident **answered `hello`** — and answering `hello` is what being in-epoch *means*. So the banner may honestly certify it, and the front relays so the parent can do its job. Handing the front the parent's policy instead looked right for about an hour and was measurably wrong: it went terminal on daemons the parent knows exactly how to drain and replace, and it took four of drishti's own e2e lanes down with it, because the front refused before the parent ever saw the resident. In production it would have made the whole skew → degraded → renew path unreachable.

That the two contract-skew fixtures now run the **production front verbatim** — no override, no fixture-specific arm — is the small proof that the layering is right.

### The boot barrier grew a second observation point

`withAgentBootBarrier` turns "the agent died with a fatal stderr line" into a terminal `ConnectError` so a planted 0755 state dir gives up immediately with the agent's own sentence, instead of retrying forever as "host unreachable".

It used to have exactly one place to observe that, because `sshConnector` returned a `Connection` unconditionally — it spliced the child's stdio into an RPC client without asking the peer anything — so a boot refusal could only ever surface downstream as `conn.closed`. The readiness gate moves that observation **earlier**: the connector now waits for the banner, races it against the child's own exit, and a child that dies without greeting makes the connector *throw* rather than return.

So the barrier catches the throw and applies the same classification. This is not a fallback — it is the same verdict at the same instant, reached by the new control path. Without it, `U3.1` regressed from "terminal, with the agent's own sentence" to "host unreachable", which is precisely the retry-forever class the gate exists to abolish. A throw with no fatal line is rethrown untouched: that is the connector's own classified verdict, and relabelling it here would take the gate's terminal budget away from the one authority that owns it.

### The policy graduated out of the app

`drishtiAgentConvergencePolicy` lived in `app/src/server/hostRegistry.ts` for the only reason that mattered at the time — the parent was the only supervisor. There are two now, and the agent's Nix closure is a *positive projection* of `packages/agent` + `packages/common`, with `packages/app` deliberately not an input. A policy left app-side is therefore not merely awkward for the front, it is **structurally invisible** to it. It moves to `drishti-common/convergence-policy` (its own export subpath, so the browser bundle never pulls daemon-supervision code), with `hostRegistry` re-exporting so no import path moved — the source of truth moved, not the API.

One typing trap is pinned in a comment for the next editor: neither policy factory may grow an explicit `: ConvergencePolicy<…>` return type. `ConnectorPolicy`, which `createConnectorDrainBudget` demands, is *narrower* on `onContractSkew`; annotating the return widens the literal and the connector arm stops compiling. `as const satisfies` proves conformance without erasing which arm was chosen.

### Closure and lockfiles

`@kolu/surface-daemon-supervisor` joins the agent's hydrated set, and `ts-pattern` joins `agent.package.json` — a **real runtime import** of that kit (`probeDaemonIdentity`, `index`), unlike `@kolu/log`, which `surface-daemon` takes as `import type` only and therefore needs no entry. The regeneration is exactly one tarball in `agent.lock` + `agent.bun.nix`, no transitive movement; the root `bun.nix` does not move at all because the app already resolved `ts-pattern` at the same version. The kit runs in the *fronting* process and never in the daemon, so it cannot change what a daemon restart loads.

Effect stays at `4.0.0-beta.103` — no vendored manifest gained a dependency this round (kolu's `surface/package.json` gained only the `./links/readiness` export entry), so `bun install` reports no changes and the pin really is just the pin. The npins hash was verified rather than assumed: the **outgoing** revision was re-prefetched first and reproduced its recorded `sha256-L8uY…` exactly, which is what makes the new `sha256-SJ/r…` trustworthy.

### The falsifier

`agent/src/epochGate.e2e.test.ts` reproduces the incident's *cause* rather than mocking its symptom. `fixtures/muteEpochDaemon.ts` is a previous-epoch daemon written as its observable behaviour — it holds the rendezvous with a real pid gate, and when something connects it **accepts and says nothing**, which is what a daemon waiting for a greeting nobody speaks any more actually does. No old release is vendored; the silence *is* the presentation.

Two e2es assert the inverse of the incident: **one takeover and one clean converge**, and then a second front that **adopts in place** — same pid, no drain, no respawn — because a re-takeover on every dial is the livelock wearing different clothes. Three pure tests pin the epoch-only policy, which is the layering decision no typechecker can see and the one most likely to be "fixed" back into a bug.

Falsified against the fix reverted: both e2es fail with the incident's own signature — `the peer accepted the pipe and sent no readiness banner within 30000ms` — the blind splice caught at the gate instead of ten seconds later as a nondescript transport death. The three policy tests correctly keep passing, since they do not depend on the front's wiring.

## Effect `4.0.0-beta.103`

kolu moved the whole stack from beta.102 to beta.103, **including the seven vendored `@kolu/surface*` literals this repo hydrates out of the Nix store**. A hydrated source and its consumer must agree on ONE copy of `effect` — two module instances mean `_tag` narrowing silently stops recognising anything — so drishti's own pins move in the same commit as the pin that brings the new sources in.

Six authored sites carry the literal here: the root `overrides`, the three package manifests, the agent's projected `agent.package.json`, and **the heredoc inside `scripts/regenerate-agent-deps.sh`** that reconstructs the agent-only workspace. The last one is the one that hides: leave it and the next regeneration silently re-pins the projection back to beta.102. Both lock/nix pairs regenerate, and the delta is exactly the three effect tarballs (`effect`, `@effect/platform-node`, `@effect/platform-node-shared`) — a url + integrity line each, no transitive dependency moved.

The npins hash was verified rather than assumed: the **outgoing** revision was re-prefetched first and reproduced its recorded `sha256-3Gpv…` exactly, which is what makes the new `sha256-L8uY…` trustworthy. `branch` also goes back to `effect` — `osfacts-effect` was a short-lived kolu branch that has since merged and been deleted upstream, so the field named a ref `npins update` could no longer follow.

**No source adaptation was needed**, and that is a swept result, not a hope. beta.103's removals were grepped for across every `.ts`/`.tsx` in `packages/`: no `SchemaIssue` / `getActual` / `Schema.redact`, no `Schema.UnknownFromJsonString`, no `SchemaMultiDocument`, no `Context.mutate` or `Context.getReferenceUnsafe`, and the single `Schema.Record` call site (`main.ts`'s agent-drv map) is the plain two-argument form the `keyValueCombiner` removal leaves alone. Nothing here touches Effect's `Clock`, so the monotonic/wall-clock split lands on nobody. The microtask-dispatch and `Atom` fixes are the ones that could have bitten — this repo runs Effect under **bun** — and the async/reactive suites are green: 375 pass · 5 skip · 0 fail across 54 files locally, and the full daemon-gated e2e lane clean on both platforms in CI.

## What the review round changed here

Eight kolu commits, and only one of them is visible from this repo.

**`pumpRemoteSurface` gained an exit case.** A mirror that ended while its link was still *alive* — every subscription settled — used to park `cursor.next()` forever: a pump silently serving nothing behind cleared holders, waiting for a spawn that was never coming. Each mirror end now races that wait against a `system.live` round-trip on the client that just ended; a far end that still answers ends the pump instead. drishti rides this pump at exactly one site — `router.ts`'s agent→parent bridge — and needs no change: the exit is a plain `return` from a `Promise<void>` that drishti already discards, no test asserts the pump parks, and the exit line does reach the parent's log because drishti passes `log`.

It is worth naming the residual rather than pretending the absorption is total. drishti `void`s that promise, and this is the **first** way the pump can resolve with the session still alive and still reading `connected` — the pump is deliberately not the authority for link health (`serveHostMap` + `session.onState` is). So a bridge that ended this way would leave a green host chip over frozen data. No test reaches it, the drain race that could reach it is unproven in either direction, and what the UI *should* say in that state is a product call rather than something to smuggle into a pin bump — so it is flagged here, not fixed here.

**The other seven are kolu-internal** (a preference-ladder fix, a typing-echo bench port, a client session-restore race and the follow-up that bounds its wait, comment-only docblock edits). The newest of them, `41d517754` — *the restore seed's wait for the answered tile is bounded* — is `packages/client` only, and drishti vendors none of `client`: the hydrated set is `surface`, `surface-remote`, `surface-map`, `shell-quote`, `log`, `surface-app`, `solid-pwa-install`, `surface-daemon`, `surface-daemon-supervisor`. No vendored manifest moved across the whole round — only `packages/server` gained a dependency and drishti vendors none of `server` either — so `bun install` reports no changes, and `bun.lock`, `bun.nix`, the agent projection (`agent.lock` / `agent.bun.nix`) and the three effect-override sites are all untouched. The re-pin is the pin.

## The scanner this repo never had

kolu's round also grew governance for the exact silent-failure class this whole campaign is about (#2101 B1): `await` on a member face compiles, resolves **with the Effect object**, and never touches the wire. drishti had *adopted* that discipline during campaign 2 but never made it **red** — the rule is written in prose three times (`wire.ts`'s `runCall`, `dialDaemon.testlib.ts`'s `unary()`, `hostRegistry.ts`'s note that the drain is an Effect value) and enforced by nothing. It held only because two well-written confinement helpers existed and everybody used them.

A **verbatim** copy of kolu's scanner would have been worse than no scanner, because it would have looked guarded while passing vacuously over the entire client tier. kolu's pattern requires a literal `.surface.` / `.procedures.` path segment; drishti spells its faces as **functions** — `hostRpc(host)`, `adminRpc()`, `hostStreams(host)`, `hostCollections(host)` — so `await adminRpc().hosts.add(x)`, the likeliest dodge in this repo, scores clean under it. The twin teaches the accessor call to *be* a face, and takes the round's two new dodges with it (the alias `const verb = …; await verb(x)`, the stored description `const p = …(x); await p`) plus the outright ban on an `Effect.run*` that is named but never called.

The sweep it pins is **clean** — all 121 `.ts`/`.tsx` files under `packages/`, zero hits in every category, so this fixes nothing and *prevents* something. Its own falsifiability is asserted in-file: each clause catches its dodge, each legitimate spelling (`runCall(...)`, `Effect.runPromise(...)`, `yield*`) survives on the nesting parens that make the scan possible at all, and a file-count floor separates a clean tree from a walk that found nothing. `runCall` is deliberately *not* banned as an uncalled reference — a bare identifier would flag its own declaration and every import of it.

## What campaign 2 added, and the one hazard it left behind

The second campaign flipped the member face to **Effect-only** (`UnaryProcedure` / `BoundProcedure` / `safe` / `isDefinedError` deleted, the `.effect` nesting folded onto `.surface`), flipped the whole supervisor face to Effect, and took `hono` out of `@kolu/surface-app`. Most of that the compiler catches.

**One class of it the compiler cannot.** An `Effect` is a *description*: `await`ing one compiles, resolves **with the Effect object**, and never touches the wire. kolu bit on this nine times, including a test that had quietly disabled the drain it existed to prove. In this repo the exposure was not the call sites — it was the two places that typed a face **as** a Promise shape through an `as unknown as` cast. Both were *true* at the old pin and became lies at this one, with nothing to catch them:

- `hostRegistry.ts`'s `ControlProbeClient` is now an **alias** of the framework's own `ControlCoreProbeClient`, not a hand-written twin. A twin restates a shape that moves whenever the frozen fragment does — and it just did.
- `dialDaemon.testlib.ts` cast three faces to Promise signatures. It **runs** the Effects now, behind one named `unary()` helper, so `await client.control.hello()` at every call site is honest again. Without this the agent's drain/adopt e2e would have passed while dispatching nothing.

The two source-grep tests that pinned `drainPersistFailureOf(await active.drain())` now pin the composed spelling **and negatively pin the awaited one**.

## The exit oracle lost its `AbortSignal`

`awaitExit` is an `Effect<void>` now, and `awaitExitViaProcessOracle` with it. The signal had exactly one job — tell a poll-shaped wait to stop once the drain ceiling won — and `convergeAdmit` forks the oracle into a scope it closes on every path, so the wait is **interrupted** rather than notified. Interruption is not refusable; an abandoned `AbortController` was. drishti's own drain is an Effect **value**, like the framework's `fireDrain`, so it cannot be `await`ed into a no-op either.

## hono is gone

`installSurfaceApp(app, opts)` became `surfaceAppLayer(opts)` — an `HttpRouter` **layer**, not an installer — so the parent's whole HTTP app moves onto `effect/unstable/http`. It owns its `http.Server` and hands the `request` event an Effect handler, which is precisely what leaves the `upgrade` event ours: the stale-tab gate and the ws reaper have to run in front of dispatch. Registration order stops meaning anything (find-my-way ranks by specificity), so the two ordering comments go with the framework instance. `hono`, `@hono/node-server` and — the reactor runs on Effect's `Atom` now — `@preact/signals-core` leave all three manifests.

## Three findings that shaped the work

**The one honest redesign is the drain.** drishti used to smuggle a failed final ring-write out of the frozen control-core drain as `ORPCError("DRISHTI_PERSIST_FAILED")`. That channel declares no error and no output, and the new epoch is explicit that a rejecting `onDrain` is a **defect** — "a daemon whose drain hook throws is broken, not busy". But a full disk is not a broken daemon: the drain *worked*, and its final write did not land. That is a verdict about a **successful call**, so it needs a value channel, not an error one.

So drishti declares its own: `daemon.ring.drain`, on a new sibling beside the frozen fragment, answering `{ persisted, error? }`. A **sibling** and not an app member, because the app surface is re-served verbatim to the browser — a drain verb there would hand every tab the authority to stop any host's daemon. Both verbs drive one latched `drainNow()`, so a generic supervisor speaking only the frozen fragment still gets a correct final write (it just does not learn the verdict), and calling both flushes exactly once. `captureDrainPersistFailure` — an error-narrowing function with a cross-realm plain-object arm — is deleted outright; there is no error to narrow any more.

**The best win is a deletion.** `admin-router.ts` hand-built a server-only *widened* oRPC contract and re-adapted two finalized routers through it with an `as any` splice — the most delicate ~40 lines in the repo, and the reason a host-map subscription could 404 while everything else stayed green. It existed because an oRPC contract was a nested matcher tree. On a flat tag namespace a tag carries its own route, so the splice is a record merge; `adminContract` disappears as a name and as a concept. What replaces the deleted matcher-tree reasoning is a route-set assertion in both directions, because on this wire **the tag set is the contract**.

**A lease stopped being a `finally`, and that mattered.** `isIdle` — the oracle a 60-minute daemon idle-exit hangs off — counts live `metricHistory` subscriptions, and used to decrement in the generator's `finally`. Under `Stream` that leaks: the framework's AsyncIterable bridge deliberately never calls `.return()` on a producer (awaiting it deadlocks a generator parked at an `await`), so the `finally` does not run when a consumer walks away. It is an `Effect.acquireRelease` on the stream's own scope now. A leaked lease would have kept a forgotten daemon alive forever.

## Spikes, run before committing to the plan

| Spike | Verdict |
| --- | --- |
| `@effect/platform-node` transports **under bun** | **PASS, and re-run at the final pin.** All four legs round-trip a unary call (including a `withDecodingDefaultKey` absent key), a stream, and a cell: `serveOverUnixSocket` ↔ `unixSocketLink` in-process; a **real `bun` child** serving over `serveOverStdio` ↔ `stdioLink`; and the `frontDaemonOverStdio` byte splice (stdio client → unix server). Re-running it at `3c631446d` needed exactly one change — the unary leg had to *run* its Effect rather than `await` it, which is the hazard above showing up in the spike itself. |
| TS **5.8** vs the ~40-tag spec-derived face | **PASS.** The whole workspace typechecks under the repo's own `typescript@^5.8` (5.9.3) with zero errors and no TS2590. No bump to `typescript@7` needed. |

One pre-existing **bun** quirk found and worth recording, unrelated to Effect: under `bun test` (1.3.10) a *handled* `connect ENOENT` on a unix path escapes as an uncaught error and fails the test — **unless** a `net.Server` has already been created in that process. It reproduces identically on `master` with the old oRPC pin, so it is bun's `node:net` client path, not ours. It never bites drishti in practice (the daemon serves in forked children, and dials target live sockets).

## The #17 mapping table as law

Every wire and disk schema moved under it: `z.enum` → `Schema.Literals`, `z.tuple` → `Schema.Tuple`, `z.discriminatedUnion("kind", …)` → `Schema.Union` (**never** `TaggedUnion` — that renames the discriminant to `_tag` and changes the bytes), `.optional()` → `Schema.optionalKey`, `.default("TERM")` → `Schema.withDecodingDefaultKey`.

The two landmines this repo has sit in the same procedure (`process.kill`'s defaulted `signal` and optional `error`), so they are pinned on the **encoded JSON string** — a decode-equality test would pass happily while `"error":null` started appearing on every successful kill. Same for all four `MetricHistoryMessage` arms, the alerts cell, and `history.ring.json`, where the fixtures also pin that an absent `alerts` key stays absent and that an explicit `null` is *corrupt* — precisely the divergence `Schema.optional` would have introduced.

## Also in here

- **The dial is one wire with three faces.** `sshConnector` takes the surface as a *value* now (Effect RPC cannot mint a client from a type). drishti dials with the app sibling's spec/prefix carrying the combined group, and builds the control-core and daemon faces over the same `Connection.dispatch`. `scopeSibling(client, "app")` is deleted — the connector's own client already *is* the app-scoped one, so the WeakMap is keyed by the very value `admit` receives.
- **The client bootstrap is async.** `connectSurfaces` returns a promise, so `wire.ts` awaits it at module scope: every consumer's import stays synchronous-looking instead of ~40 call sites learning the wire might not exist yet. The host map's tags ride `extraGroups` — the client twin of the server's merge, because Effect RPC resolves a call's schemas by tag lookup and a tag outside the group cannot be dispatched at all.
- **Both `AbortController`s in `App.tsx` are gone.** `unenrolledStreamCall` returns a `Stream` and `createSubscription` owns teardown as a fiber interrupt — the controllers *were* that interrupt, spelled by hand. partysocket's 60s cold-start `connectionTimeout` goes too, and its absence is the point: it existed to stop a 4s default from abandoning a dial while `nix copy` ran, and the Effect link never times an attempt out at all.
- **`daemonStatus.ts` gains the sixth `unspeakable-protocol` arm** — a daemon from the *previous* protocol epoch whose first frame this supervisor cannot decode. Both exhaustive switches grew a case, which is exactly what the closed union is for.
- **The client has ONE run edge**, `wire.ts`'s `runCall`. Solid's event handlers and store writers are synchronous callbacks with no Effect slot to compose into, so the seam is real; naming it once keeps it countable instead of letting `Effect.runPromise` spread across seven call sites.
- Dependencies: `@orpc/*`, `zod`, `partysocket`, `hono`, `@hono/node-server` and `@preact/signals-core` are gone; `effect` + `@effect/platform-node` land at the **literal** `4.0.0-beta.103` the hydrated sources declare, in all three override sites in lockstep (root manifest, agent projection manifest, and the heredoc inside `regenerate-agent-deps.sh`) — a fourth copy that drifts is how the agent projection silently diverges.

## Evidence

Beyond CI, the built binary was driven end to end in a real browser — and **re-driven at the final pin**, after the HTTP stack moved off hono: 519 processes streaming through the collection `deltas` verb, 16 CPU cores, 3 NICs, 8 unclaimed listeners, live load/mem/swap/disk, and the metric-history chart — every primitive class on the surface, over the new wire, with a clean console. The `identity.info` probe resolving on both the `SRV` and `CLIENT` footers is the surface-app sibling answering over the same socket.

The freshness contract was curled straight off the new Effect HTTP layer, under bun: `/` and `/t/abc` and a missing `/foo.png` → the `no-store` shell; `/assets/<hashed>.js` → `public, max-age=31536000, immutable`; `/assets/gone-999.js` → **404 + `no-store`, never the shell**; `/sw.js` → `no-cache, must-revalidate` + `text/javascript`; `/manifest.webmanifest` → `application/manifest+json`; and a request carrying the response's own `ETag` back still gets **200, never a 304** — the suppression that keeps a Nix-store epoch mtime from replaying a stale shell (kolu#1319).

The host-map fold was also exercised directly over a real WebSocket against the built server (`surface/hosts/entries/keys`, a folded per-host `system` cell, and `metricHistory`) — the exact tag class the deleted widening seam used to 404 on.


srid:
## Evidence

The **built** binary (`nix build .#default`), driven in a real browser against a live `localhost` host — the whole chain on the new wire: browser → `websocketLink` → the parent's `serveSurfaceSocket` → the host-map fold → `pumpRemoteSurface` → `sshConnector` → the agent daemon over stdio + unix socket → osfacts.

**Fleet view** — the host map's `entries` membership and the folded per-host `system` cell:

fleet

**Host view** — every primitive class on the surface at once: 447 processes through the collection `deltas` verb, 16 CPU cores, 3 NICs, 8 unclaimed listeners, the `alerts` cell, and the `metricHistory` stream:

host

Console: clean (one pre-existing SolidJS "computations created outside a `createRoot`" warning from `wire.ts`'s module-scope subscriptions — unchanged by this PR).

The host-map fold was also exercised directly over a real WebSocket against the same built server, which is the tag class the deleted widening seam used to 404 on:

```
[smoke] hosts/entries/keys        -> Some(["localhost"])
[smoke] hosts/localhost system    -> Some({"loadAvg":[0,0,0],"cpuPct":0,…})
[smoke] hosts/localhost metricHistory -> Some({"kind":"snapshot","samples":[]})
```

### Spike A — `@effect/platform-node` transports under bun

Run against the pinned kolu sources before any wave was committed to. All four legs round-trip a unary call (including a `withDecodingDefaultKey` **absent** key and an `optionalKey` absent field), a stream, and a cell:

| Leg | Result |
| --- | --- |
| `serveOverUnixSocket` ↔ `unixSocketLink`, in-process | pass |
| a **real `bun` child** serving `serveOverStdio` ↔ `stdioLink` | pass |
| `frontDaemonOverStdio`'s byte splice — stdio client → unix server | pass |
| a dead path rejects the dial (`ENOENT`), preserving the "nothing is serving here" verdict | pass |

PR #133, opened 2026-08-04
feat(agent): adopt the Effect-native osfacts client

srid:
osfacts' TypeScript client went Effect-native in [juspay/osfacts#5](https://github.com/juspay/osfacts/pull/5): every spawning verb now hands back an `Effect.Effect<_, OsfactsClientError>` instead of a Promise, and `OsfactsClientError` became a union of three tagged classes rather than one class with a `kind` string. drishti consumes exactly **two** of those verbs — `snapshotHost` and `host`, both in `packages/agent/src/proc.ts` — so this adoption is two run edges, their test doubles, and the pin that brings the new client in.

## What changed

**`proc.ts` — two run edges, and where they are.** `readSnapshot(...)` and `readHost(...)` are now descriptions; `Effect.runPromise` runs them inside the existing cache-window guard, which is what keeps "one osfacts subprocess per fact family per window" true. The edges live *in* `proc.ts` rather than higher up because `ProcReader` feeds the reactor's `source({ read })`, whose `read` is `() => Promise<T>` by design — pushing the Effects up would only relocate the run, not remove it, and would leave two Promise faces on one reader. That is the same call kolu made for padi's port sampler in the paired PR, and the same one `hostRegistry.ts` already documents at the `Admit` seam it doesn't own.

**`hostFromOsfacts` stays *outside* the host Effect, deliberately.** It throws `OsfactsSourceError`, and `readSourceErrors` recovers that by reading a marker out of the error's own message. Folded in as an `Effect.map`, the throw would become a **defect**, reach the caller wrapped, and the marker read would go blind. The comment at the edge says so.

**The seam types needed no edit — which is the point.** `SnapshotHost` / `ReadHost` are `typeof snapshotHost` / `typeof host`, so the upstream API change arrived as a type error at every call site rather than as doubles that had quietly drifted from the thing they double. The doubles are now `Effect.sync`, which upgrades the two counting tests from counting *descriptions built* to counting *spawns*.

**One falsified pin.** `hands a failed osfacts read through the run edge as the client's own tagged error` asserts **identity** (`toBe`, not `toBeInstanceOf`): a runner that wrapped the failure in a fiber-failure envelope would silently break `readSourceErrors`' marker recovery, and the rest of the suite would stay green. Made red by wrapping the rejection at the edge, then green again.

## What did *not* change, and why that was checked rather than assumed

| Touchpoint | Verdict |
|---|---|
| `processIdentity.ts` | The port kept `processIdentity` on the sync island. Compiles as-is. |
| `parseHostOutput` / `parseSnapshotOutput` in `proc.test.ts` | Still sync-throwing. Call shapes unchanged. |
| Supervisor seam | drishti uses `convergeAdmit` and `probeDaemonIdentityFrom` — neither takes an osfacts reader. drishti never calls `createEndpoint`, so the two flipped seams (`ReadSocketHolders`, `ReadProcessIdentityAsync`) have no drishti consumer. |
| Error vocabulary | No `instanceof OsfactsClientError` and no `.kind` read existed anywhere in drishti. |
| `@kolu/surface-daemon`'s pid gate | Untouched upstream — it takes the *sync* `ProcessIdentity`, which is exactly what `processIdentity.ts` supplies. |

## The pin

kolu at `7d58e57f0` (branch `osfacts-effect`, the head of juspay/kolu#2103), which pins osfacts at `86091cf6`. drishti reads osfacts through **kolu's own** npins pin (`nix/overlay.nix`), so one bump moves both the binary and its client together.

The `branch` field moves to `osfacts-effect` alongside the revision. Left at `effect` it would be a lie that a future `npins update` would "fix" by silently yanking the pin back off the adoption.

## Pair relationships and merge order

This is one of four PRs that only typecheck together:

1. [juspay/kolu#2101](https://github.com/juspay/kolu/pull/2101) — the Effect 4 migration these all stand on
2. [juspay/osfacts#5](https://github.com/juspay/osfacts/pull/5) — the client port itself (`86091cf6`, green)
3. [juspay/kolu#2103](https://github.com/juspay/kolu/pull/2103) — kolu's adoption: bumps its osfacts pin and flips the supervisor's last two Promise seams
4. **this PR**, stacked on [#132](https://github.com/srid/drishti/pull/132) (base `effect`)

Merge in that order, each step re-pinning to the previous one's **merge commit** rather than its branch head.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -7,11 +7,11 @@
         "owner": "juspay",
         "repo": "kolu"
       },
-      "branch": "effect",
+      "branch": "osfacts-effect",
       "submodules": false,
-      "revision": "41d517754737ad2b97e72719e276739102d749b9",
-      "url": "https://github.com/juspay/kolu/archive/41d517754737ad2b97e72719e276739102d749b9.tar.gz",
-      "hash": "sha256-nZPp6CF7ZHR9Ev+Fjy2tRi6kwhae+jHJUqJVo0NLvQE="
+      "revision": "c72da36cd80b1abb016209792708ccd5b02d5bf9",
+      "url": "https://github.com/juspay/kolu/archive/c72da36cd80b1abb016209792708ccd5b02d5bf9.tar.gz",
+      "hash": "sha256-3Gpv4Gd2R9dp1XQLh/EUOUJlwVfhDg9mkyNRmg67Rnk="
     },
     "nixpkgs": {
       "type": "Git",
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -9,8 +9,8 @@
       },
       "branch": "osfacts-effect",
       "submodules": false,
-      "revision": "c72da36cd80b1abb016209792708ccd5b02d5bf9",
-      "url": "https://github.com/juspay/kolu/archive/c72da36cd80b1abb016209792708ccd5b02d5bf9.tar.gz",
+      "revision": "7d58e57f08a2b8a3c33c230b5f518ee05a928a76",
+      "url": "https://github.com/juspay/kolu/archive/7d58e57f08a2b8a3c33c230b5f518ee05a928a76.tar.gz",
       "hash": "sha256-3Gpv4Gd2R9dp1XQLh/EUOUJlwVfhDg9mkyNRmg67Rnk="
     },
     "nixpkgs": {
--- a/packages/agent/src/proc.test.ts
+++ b/packages/agent/src/proc.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from "bun:test";
+import { Effect } from "effect";
 import { osfactsSourceStatus } from "drishti-common/source-errors";
-import { parseHostOutput, parseSnapshotOutput } from "osfacts-client";
+import {
+  OsfactsSpawnError,
+  parseHostOutput,
+  parseSnapshotOutput,
+} from "osfacts-client";
 import {
   computeCpuUsage,
   computeNetThroughput,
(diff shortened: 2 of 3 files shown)


srid:
## CI — green on both platforms

odu run **`73c4f98#1`** · 2 lanes · **24 ok · 0 failed · 0 errored · 0 skipped**

| platform | host | nodes |
|---|---|---|
| `x86_64-linux` | `kolu-ci-1` | 12 ✔ |
| `aarch64-darwin` | `ci@petit` | 12 ✔ |

All 22 required `<recipe>@<platform>` contexts from `odu protect --dry-run --platform x86_64-linux --platform aarch64-darwin` are green, plus `_ci-setup@<platform>` and both GHA `build-and-push` jobs. Nothing was retried and no opt-out flag was used — strict, posting, snapshotting.

The slowest node on each lane: `ci::test-daemon@x86_64-linux` at 1m5s (the bun suite with `KOLU_DAEMON_TESTS=1`) and `ci::drv-stability@x86_64-linux` at 31s. drishti's pipeline has no `e2e` recipe, so there is no e2e duration to report.

`ci::test-daemon@x86_64-linux` is the node that actually exercises this change: 375 pass / 5 skip / 0 fail, including the new run-edge pin.

srid:
Folded into srid/drishti#132 by maintainer decision — its `effect` branch fast-forwarded to this head `73c4f98`, so #132 now carries the osfacts-client adoption (CI green at this exact SHA carries over: odu run `73c4f98#1`, 24 ok both platforms). One PR per repo.

PR #142, opened 2026-08-15
Pin kolu to juspay/kolu#2170 (per-face `expose`) — the surface-rule pair PR

srid:
The pair PR [`.claude/rules/surface.md`](https://github.com/juspay/kolu/blob/master/.claude/rules/surface.md) requires for an API-facing change to `@kolu/surface` / `@kolu/surface-app`. Pairs with **juspay/kolu#2170**.

kolu#2170 gives each serving face its own default-deny `expose` allowlist — the same `ExposeMap` shape `serveSurfaceAsMcp` already took, so a verb can be reachable over a trusted unix socket and refused to browsers. `serveSurfaceApp` and `serveOverUnixSocket` gain an optional `expose?: FaceExposure`, and a new `@kolu/surface/expose` subpath exports the shape (`ExposeMap` / `ToolExposure`, moved there from `@kolu/surface-mcp`), the single key grammar every face reads a map through (`classifyExpose`), the binders (`exposeFace` / `exposeFaces`) and the filter (`restrictHandlers`).

**Nothing drishti imports changed shape.** The addition is a new subpath plus a new *optional* option; drishti imports none of the moved names, and its listener takes the granular `serveSurfaceSocket` seam, which is unchanged and deliberately takes no map (a hand-built serve path that wanted a gate would call `restrictHandlers` itself — drishti does not today).

So this is the gate rather than a migration: it exists to prove drishti still builds and passes CI against the new surface API. Pinned to kolu#2170's **final** HEAD (`6180b074`), after its review gauntlet and CI fix-commits — the rule is explicit that a green pinned to a pre-gauntlet SHA is stale.

`just typecheck` is green against this pin locally (common + agent + app).

**Awaiting srid** — do not merge; the kolu side is reviewed and merged by srid personally, and this should follow it.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -7,11 +7,11 @@
         "owner": "juspay",
         "repo": "kolu"
       },
-      "branch": "master",
+      "branch": "surface-per-face-expose",
       "submodules": false,
-      "revision": "dad81604eaaf66ec5ff59942973865f62bcddefd",
-      "url": "https://github.com/juspay/kolu/archive/dad81604eaaf66ec5ff59942973865f62bcddefd.tar.gz",
-      "hash": "sha256-gIM4+eC6CTze9hHVAR7oEI2zyJXvd+h8+bUVQgtpweQ="
+      "revision": "6180b07441510907e98584a4288fa15a68aacb27",
+      "url": "https://github.com/juspay/kolu/archive/6180b07441510907e98584a4288fa15a68aacb27.tar.gz",
+      "hash": "sha256-sLHq5gE6FR57GmYHc6me9SL1Qs3yNbH+IE/rgtAofBw="
     },
     "nixpkgs": {
       "type": "Git",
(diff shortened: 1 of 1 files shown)


srid:
Re-pinned to kolu **master** — juspay/kolu#2170 merged.

#2170 was **squash**-merged (`bfb4eb56`), so the commit this PR was pinned to (`3a2d8609`) is not an ancestor of kolu master and its branch is gone — a dangling pin that resolves only while GitHub still serves the orphan archive. Back to steady state: `branch: master`, revision `bfb4eb56`.

The tarball hash is **byte-identical** to `3a2d8609`'s, which is the useful cross-check: the squash carried the reviewed tree over unchanged, so this is the same code the gate was already green against — now reachable from master. `just typecheck` green against the new pin; CI re-running on this head (`787d30f`).

Ready to merge once its checks are green. Not merging it myself.

PR #147, opened 2026-08-26
Pin kolu to juspay/kolu#2217 (daemon leaf entries) — the surface-rule pair PR

srid:
The pair PR [`.claude/rules/surface.md`](https://github.com/juspay/kolu/blob/master/packages/AGENTS.md) requires for an API-facing change to `@kolu/surface-daemon` and `@kolu/surface-daemon-supervisor`.

Pairs with **[juspay/kolu#2217](https://github.com/juspay/kolu/pull/2217)**, pinned at its HEAD `aa19a1106`.

## What changed upstream

Each daemon package grew a narrow entry beside its barrel:

| leaf | carries | no longer compiles |
| --- | --- | --- |
| `@kolu/surface-daemon-supervisor/dial` | `dialSocket`, `DaemonContractSkewError`, `isContractSkewError`, `DaemonConnection` | `endpoint.ts` — the drivers, the convergence probe, and `osfacts-client` |
| `@kolu/surface-daemon/home` | `resolveDaemonHome` + the home path algebra | `daemonMain`, `daemonProcessMain`, the process-signal tier |
| `@kolu/surface-daemon/lifetime` | `DaemonLifetime`, `DaemonLifetimeInfo`, `lifetimeInfo` | the same |

The point is a browser-safe consumer of `connectPadi` that stops compiling a daemon runtime — kolu's own hydrate guard now asserts that **no bare daemon barrel** is in `@kolu/padi-client`'s import closure, where it previously recorded both as a known cost.

## Why this is the gate, not a migration

**Nothing drishti imports changed shape.** Both barrels' exported-symbol lists are byte-identical: the moved symbols are imported by their old homes and re-exported from them, so `endpoint.ts` and `daemonMain.ts` still export everything they did. The only public delta is three **added** `exports` entries.

Drishti reaches the barrels and one testlib, and names no deep path into either split module:

```
17  from "@kolu/surface-daemon"
 4  from "@kolu/surface-daemon-supervisor"
 3  from "@kolu/surface-daemon/upgrade-window.testlib"
```

So this PR exists to prove drishti still builds and passes CI against the new API — which is exactly what the rule asks for when the change is additive.

One thing worth stating rather than leaving implied: **drishti still hydrates `osfacts-client`, and still should.** kolu#2217 retires that second pin for consumers that reach `@kolu/padi-client/dial`; drishti is a direct supervisor-**barrel** consumer, so `endpoint.ts` — and with it the osfacts process-identity vocabulary — is still legitimately in its program.

## Evidence

- `just typecheck` — green across `common`, `agent`, `app`.
- `just test` — **387 pass / 7 skip / 0 fail**, 1064 assertions, 394 tests across 57 files.

The pin will be re-pointed at kolu master once #2217 merges. Per the rule, it will also be bumped to #2217's final post-gauntlet HEAD and re-validated before either side is called done.

https://claude.ai/code/session_01UWaE2b2irwGkqt54XV7Dzd

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -7,10 +7,10 @@
         "owner": "juspay",
         "repo": "kolu"
       },
-      "branch": "dock-row-export",
+      "branch": "master",
       "submodules": false,
-      "revision": "de8fcad8d56a445c8a71d7fa65b7c565c1fa4299",
-      "url": "https://github.com/juspay/kolu/archive/de8fcad8d56a445c8a71d7fa65b7c565c1fa4299.tar.gz",
+      "revision": "f9ace75e6dc25a2bbada31bd1daacf0c28405934",
+      "url": "https://github.com/juspay/kolu/archive/f9ace75e6dc25a2bbada31bd1daacf0c28405934.tar.gz",
       "hash": "sha256-+vIxBhOFvbMNgaLOBzQKGizc8QU7cm76I0+MoWL5I48="
     },
     "nixpkgs": {
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -9,9 +9,9 @@
       },
       "branch": "dock-row-export",
       "submodules": false,
-      "revision": "49db20651caafec6b2f00d727e1c55260ef76371",
-      "url": "https://github.com/juspay/kolu/archive/49db20651caafec6b2f00d727e1c55260ef76371.tar.gz",
-      "hash": "sha256-VDesZpcKCsXDHQ05HGrjh0RpjFEDgj3ITW+wyTGYMBM="
+      "revision": "de8fcad8d56a445c8a71d7fa65b7c565c1fa4299",
+      "url": "https://github.com/juspay/kolu/archive/de8fcad8d56a445c8a71d7fa65b7c565c1fa4299.tar.gz",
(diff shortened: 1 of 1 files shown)


srid:
### Now riding merged master

kolu#2217 merged as [`f9ace75e`](https://github.com/juspay/kolu/commit/f9ace75e6dc25a2bbada31bd1daacf0c28405934). This pair is re-pinned from the PR branch head (`de8fcad8d`, `dock-row-export`) to that commit, and the pin's `branch` is back to `master` — it pointed at the PR branch only for the duration of the pair.

**The pin hash did not change**: `sha256-+vIxBhOFvbMNgaLOBzQKGizc8QU7cm76I0+MoWL5I48=` before and after. The squash landed on a master that had not moved since the fork point, so the tree at `f9ace75e` is byte-identical to the tree this pair was already validated against. The revision label changed; nothing drishti compiles did.

Re-validated at the merged sha regardless — `just typecheck` green across `common`/`agent`/`app`, `just test` **387 pass / 7 skip / 0 fail**.

The gate verdict is unchanged from the earlier rounds: both daemon barrels' exported-symbol lists remain byte-identical (the moved symbols are re-exported from their old homes), drishti names no deep path into either split module, and the contract-5.5 fields are optional on schemas drishti does not decode — it consumes `@kolu/surface*`, not padi's terminal vocabulary.

Ready when you are; this merges on your word.

